PR #4, opened 2016-10-17
Fix mono test

gmartin7:
Ubuntu Precise does not have 'az' culture. Trusty does.
This test will be ignored on Precise, but will pass correctly on Trusty.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/4)

### DIFF
--- a/src/L10NSharpTests/LocalizationManagerTests.cs
+++ b/src/L10NSharpTests/LocalizationManagerTests.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Windows.Forms;
@@ -200,6 +201,10 @@ namespace L10NSharp.Tests
 		[Test]
 		public void GetUiLanguages_AzeriHasHackedNativeName()
 		{
+			// Check if the OS includes 'az' culture - Ubuntu 12.04 Precise doesn't
+			if (!CultureInfo.GetCultures(CultureTypes.NeutralCultures).Select(c => c.Name == "az").Any())
+				Assert.Ignore("Test requires the availability of the 'az' culture");
+
 			var cultures = LocalizationManager.GetUILanguages(false);
 			Assert.AreEqual("Azərbaycan dili", cultures.Where(c => c.Name == "az").Select(c => c.NativeName).FirstOrDefault());
 		}


andrew-polk:
Reviewed 1 of 1 files at r1.

---

*[src/L10NSharpTests/LocalizationManagerTests.cs, line 205 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/4#-KUIlLiMQ_J3ym0VYMpJ:-KUIlLiMQ_J3ym0VYMpK:bmkmpj6) ([raw file](https://github.com/sillsdev/l10nsharp/blob/b5899cfd58c5d8dd9ec02b4b36941054423880a0/src/L10NSharpTests/LocalizationManagerTests.cs#L205)):*
> ```Smalltalk
> 		{
> 			// Check if the OS includes 'az' culture - Ubuntu 12.04 Precise doesn't
> 			if (!CultureInfo.GetCultures(CultureTypes.NeutralCultures).Select(c => c.Name == "az").Any())
> ```

Seems okay in general...
But makes me wonder if there is another culture which could be tested on all systems to accomplish the original intent of the test.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/4)*

gmartin7:

---

*[src/L10NSharpTests/LocalizationManagerTests.cs, line 205 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/4#-KUIlLiMQ_J3ym0VYMpJ:-KUImMk6Q4yV657E158V:b-fatolc) ([raw file](https://github.com/sillsdev/l10nsharp/blob/b5899cfd58c5d8dd9ec02b4b36941054423880a0/src/L10NSharpTests/LocalizationManagerTests.cs#L205)):*
<details><summary><i>Previously, andrew-polk wrote…</i></summary>
> Seems okay in general...
> But makes me wonder if there is another culture which could be tested on all systems to accomplish the original intent of the test.
</details>
I doubt it, since this is a just a hack to make sure that NativeName returns the 'right' value according to an expert for just this one language:
	/// This class exists only to override Microsoft's NativeName for the Azeri neutral culture.
	/// According to Ken Keyes the NativeName on the 'az-Latn' culture is correct, while the
	/// NativeName on the 'az' culture is incorrect. So we do a swifty swap.
	/// </summary>
	public class L10NCultureInfo: CultureInfo
and
			// The Windows .Net runtime returns 'Azərbaycan dili (Azərbaycan)' for az-Latn, and
			// something totally different for az.
			// The Mono runtime returns "azərbaycan" for both az and az-Latn.
			// So we just set this to what we "know" is the right value for az.
			if (name == "az")
				_nativeName = "Azərbaycan dili";

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/4)*

andrew-polk:
:lgtm:

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/4#-:-KUInAPcNMh_qCwxMPsl:bnfp4nl)*

PR #6, opened 2016-10-17
Update to .Net 4.6.1

gmartin7:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/6)


andrew-polk:
I'm not going to claim to be able to fully analyze the changes to all those files, but I'm assuming you simply changed the .NET version for each project from 4.0 Client Framework to 4.6.1.

I think we need to change the version number with this change. At least minor, but possibly major since I think it means no application which hasn't been upgraded to 4.6.1 will be able to use it.
Not sure if that is done in code for L10NSharp or only on TC.

PR #15, opened 2017-07-05
Update L10NSharp build to download NuGet packages via msbuild

StephenMcConnel:
I copied the NuGet.targets file from BloomDesktop in case you were wondering about the changes in that file.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/15)

### DIFF
--- a/build/NuGet.targets
+++ b/build/NuGet.targets
@@ -6,6 +6,9 @@
 		<!-- Enable the restore command to run before builds -->
 		<RestorePackages Condition="  '$(RestorePackages)' == '' ">true</RestorePackages>
 
+		<!-- Property that enables building a package from a project -->
+		<BuildPackage Condition=" '$(BuildPackage)' == '' ">false</BuildPackage>
+
 		<!-- Determines if package restore consent is required to restore packages -->
 		<RequireRestoreConsent Condition=" '$(RequireRestoreConsent)' != 'false' ">false</RequireRestoreConsent>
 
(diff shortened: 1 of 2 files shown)


andrew-polk:
:lgtm:

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/15#-:-KoJ4s1xma9KGAZkf3R0:bnfp4nl)*

ermshiperete:
I'm not sure what motivated this change. Wasn't restoring packages already done before when called from the command line: `xbuild /t:Test build/L10NSharp.proj`.

This introduces the deprecated [MSBuild-integrated restore](https://docs.microsoft.com/en-us/nuget/consume-packages/package-restore#msbuild-integrated-restore) which has the disadvantage that it'll always tries to restore the packages when building. This can slow down things if you do a lot of builds and the packages didn't change. IMO it would be better to use one of the other approaches for package restore listed in the [nuget docs](https://docs.microsoft.com/en-us/nuget/consume-packages/package-restore).

StephenMcConnel:
Okay, I see that those targets with that project file do work on Linux.  But they are not documented, so I didn't know they existed, and I lost 2-3 hours of work yesterday trying to figure something out that would work.  And msbuild on Windows worked and restored the packages with building with the solution, making me think that Linux wasn't working right.   We need a consistent approach across our projects for using NuGet.  I fully understand the disadvantage of having an integrated restore that runs every time you build, but that's how Bloom has been working and it's the first project I've worked on that uses NuGet.

ermshiperete:
Sorry you lost so much time trying to get a working solution.

I'd recommend to rely on the package restore in the IDE (Visual Studio, MD, and Rider can all restore the packages when loading the solution). For the cases where we don't use an IDE (like when building from the command line) I'd import the `NuGet.targets` file into the `*.proj` file and make the `Build` target depend on `RestorePackages`.

By the way, tried to introduce this approach in Bloom ([commit d541fa27](https://github.com/BloomBooks/BloomDesktop/commit/d541fa271b730e06d63d3e14fb4d135410653c25)), but unfortunately it got later reverted.

PR #16, opened 2017-07-07
Get xliff branch working with Bloom and remove warnings

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/16)

### DIFF
--- a/src/L10NSharp/CodeReader/StringExtractor.cs
+++ b/src/L10NSharp/CodeReader/StringExtractor.cs
@@ -84,9 +84,9 @@ namespace L10NSharp.CodeReader
 							}
 						}
 					}
-					Debug.WriteLine(String.Format("DEBUG: StringExtractor.DoExtractingWork() loaded resources for {0}", type.FullName));
+					//Debug.WriteLine(String.Format("DEBUG: StringExtractor.DoExtractingWork() loaded resources for {0}", type.FullName));
 				}
-				catch (MissingManifestResourceException e)
+				catch (MissingManifestResourceException /*e*/)
 				{
 					// If it doesn't find any resources, no reason to die, we're just making a best attempt.
 					Debug.WriteLine(String.Format("DEBUG: StringExtractor.DoExtractingWork() could not load resources for {0}", type.FullName));
--- a/src/L10NSharp/LocalizedStringCache.cs
+++ b/src/L10NSharp/LocalizedStringCache.cs
@@ -24,11 +24,6 @@ namespace L10NSharp
 		// character which, when included, messes up the display of text in controls.
 		internal const string kOSRealNewline = "\n";
 
-		/// <summary>
-		/// review: I (JH) don't know what this is about
-		/// </summary>
-		internal const string kHardLineBreakReplacementProperty = "x-hardlinebreakreplacement";
-
 		private readonly string _ampersandReplacement = "|amp|";
 
 		// This is the symbol for a newline that users put in their localized text when
@@ -52,6 +47,7 @@ namespace L10NSharp
 		{
 			OwningManager = owningManager;
 			XliffDocument = CreateEmptyStringFile();
+			XliffDocument.File.Original = owningManager.Name + ".dll";
 			try
 			{
 				MergeXliffFilesIntoCache(OwningManager.XliffFilenamesToAddToCache);
@@ -69,7 +65,7 @@ namespace L10NSharp
 			if (replacement != null)
 				_ampersandReplacement = replacement;
 
-			replacement = XliffDocument.File.GetPropValue(kHardLineBreakReplacementProperty);
+			replacement = XliffDocument.File.HardLineBreakReplacement;
 			if (replacement != null)
 				s_literalNewline = replacement;
 
@@ -162,7 +158,7 @@ namespace L10NSharp
 			var xliffDoc = new XLiffDocument();
 			xliffDoc.File.SourceLang = LocalizationManager.kDefaultLang;
 			xliffDoc.File.ProductVersion = "0.0.0";
-			xliffDoc.File.Header.Note.Text = "hardlinebreakreplacement:" + s_literalNewline;
+			xliffDoc.File.HardLineBreakReplacement = s_literalNewline;
 			return xliffDoc;
 		}
 
@@ -245,7 +241,9 @@ namespace L10NSharp
 			var xliffDoc = CreateEmptyStringFile();
 			xliffDoc.File.SourceLang = langId;
 			xliffDoc.File.ProductVersion = OwningManager.AppVersion;
-			xliffDoc.File.Header.Note.Text = "hardlinebreakreplacement:" + s_literalNewline;
+			xliffDoc.File.HardLineBreakReplacement = s_literalNewline;
+			if (OwningManager != null && OwningManager.Name != null)
+				xliffDoc.File.Original = OwningManager.Name + ".dll";
 
 			foreach (var tu in XliffDocument.File.Body.TransUnits)
 			{
--- a/src/L10NSharp/TransUnitUpdater.cs
+++ b/src/L10NSharp/TransUnitUpdater.cs
@@ -25,7 +25,7 @@ namespace L10NSharp
 		internal TransUnitUpdater(XLiffDocument xliffDoc)
 		{
 			_xliffDoc = xliffDoc;
-			var replacement = _xliffDoc.File.GetPropValue(LocalizedStringCache.kHardLineBreakReplacementProperty);
+			var replacement = _xliffDoc.File.HardLineBreakReplacement;
 			if (replacement != null)
 				_literalNewline = replacement;
 		}
@@ -92,7 +92,7 @@ namespace L10NSharp
 		{
 			if (locInfo.DiscoveredDynamically && (tu.GetPropValue(LocalizedStringCache.kDiscoveredDyanmically) != "true"))
 			{
-				tu.Type = LocalizedStringCache.kDiscoveredDyanmically;
+				tu.Dynamic = true;
 				_updated = true;
 			}
 
@@ -129,7 +129,8 @@ namespace L10NSharp
 
 					_updated = true;
 					tu.RemoveVariant(tuv);
-					if (tu.Source == null && tu.Target == null)
+					if ((tu.Source == null || tu.Source.Value == null) &&
+						(tu.Target == null || tu.Target.Value == null))
 					{
 						_xliffDoc.RemoveTransUnit(tu);
 						tu = null; // so we will make a new one if needed.
--- a/src/L10NSharp/UI/CustomDropDownComboBox.cs
+++ b/src/L10NSharp/UI/CustomDropDownComboBox.cs
@@ -13,9 +13,6 @@ namespace L10NSharp.UI
 		private PopupControl m_popupCtrl;
 
 		/// ------------------------------------------------------------------------------------
-		public EventHandler PopupClosed;
-
-		/// ------------------------------------------------------------------------------------
 		/// <summary>
 		/// Initializes a new instance of the <see cref="CustomDropDownComboBox"/> class.
 		/// </summary>
@@ -98,27 +95,7 @@ namespace L10NSharp.UI
 		internal PopupControl PopupCtrl
 		{
 			get { return m_popupCtrl; }
-			set
-			{
-				if (m_popupCtrl != null)
-					m_popupCtrl.PopupClosed -= OnPopupClosed;
-
-				m_popupCtrl = value;
-
-				if (m_popupCtrl != null)
-					m_popupCtrl.PopupClosed += OnPopupClosed;
-			}
-		}
-
-		/// ------------------------------------------------------------------------------------
-		/// <summary>
-		/// Handles the VisibleChanged event of the m_popupCtrl control.
-		/// </summary>
-		/// ------------------------------------------------------------------------------------
-		protected virtual void OnPopupClosed(object sender, EventArgs e)
-		{
-			if (PopupClosed != null)
-				PopupClosed(this, e);
+			set { m_popupCtrl = value; }
 		}
 
 		/// ------------------------------------------------------------------------------------
--- a/src/L10NSharp/UI/L10NExtender.cs
+++ b/src/L10NSharp/UI/L10NExtender.cs
@@ -33,7 +33,10 @@ namespace L10NSharp.UI
 	{
 		private static HashSet<Type> s_doNotExtend;
 
+		#pragma warning disable CS0414	// field is assigned a value but its value is never used
+		// Required for Windows.Forms Class Composition Designer support
 		private Container components = null;
+		#pragma warning restore CS0414
 		private Dictionary<object, LocalizingInfo> m_extendedCtrls;
 		private LocalizationManager _manager;
 		private string _locManagerId;
--- a/src/L10NSharp/UI/ShortcutKeysDropDown.cs
+++ b/src/L10NSharp/UI/ShortcutKeysDropDown.cs
@@ -24,17 +24,6 @@ namespace L10NSharp.UI
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Handles the shortcut keys drop down closed.
-		/// </summary>
-		/// ------------------------------------------------------------------------------------
-		protected override void OnPopupClosed(object sender, EventArgs e)
-		{
-			base.OnPopupClosed(sender, e);
-			SetCurrentKeysFromEditor();
-		}
-
-		/// ------------------------------------------------------------------------------------
-		/// <summary>
 		/// Gets or sets the shortcut keys.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
--- a/src/L10NSharp/UI/Utils.cs
+++ b/src/L10NSharp/UI/Utils.cs
@@ -72,8 +72,6 @@ namespace L10NSharp.UI
 
 			return binding.GetType().GetMember(propertyName,
 				flags | BindingFlags.Instance).Length > 0;
-
-			return false;
 		}
 
 		/// ------------------------------------------------------------------------------------
--- a/src/L10NSharp/XLiffUtils/TransUnit.cs
+++ b/src/L10NSharp/XLiffUtils/TransUnit.cs
@@ -42,15 +42,15 @@ namespace L10NSharp.XLiffUtils
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Gets the type of translation unit.
+		/// Gets whether the unit is "dynamic" (discovered dynamically while the program is running).
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
-		[XmlAttribute("extype")]
-		public string Type { get; set; }
+		[XmlAttribute("dynamic", Namespace=XLiffXmlSerializationHelper.kSilNamespace), System.ComponentModel.DefaultValue(false)]
+		public bool Dynamic { get; set; }
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Gets the list of translation unit.
+		/// Gets the source (original) text (and language) of the translation unit.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		[XmlElement("source")]
@@ -58,7 +58,7 @@ namespace L10NSharp.XLiffUtils
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Gets the list of translation unit.
+		/// Gets the target (translated) text (and language) of the translation unit.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		[XmlElement("target")]
@@ -66,7 +66,7 @@ namespace L10NSharp.XLiffUtils
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Gets the list of translation unit.
+		/// Gets the list of translation notes.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
 		[XmlElement("note")]
--- a/src/L10NSharp/XLiffUtils/XLiffBody.cs
+++ b/src/L10NSharp/XLiffUtils/XLiffBody.cs
@@ -92,8 +92,6 @@ namespace L10NSharp.XLiffUtils
 		/// <summary>
 		/// When all but the last part of the id changed, this can help reunite things
 		/// </summary>
-		/// <param name="id"></param>
-		/// <returns></returns>
 		internal TransUnit GetTransUnitForOrphan(TransUnit orphan)
 		{
 			var terminalIdToMatch = LocalizedStringCache.GetTerminalIdPart(orphan.Id);
--- a/src/L10NSharp/XLiffUtils/XLiffDocument.cs
+++ b/src/L10NSharp/XLiffUtils/XLiffDocument.cs
@@ -85,9 +85,10 @@ namespace L10NSharp.XLiffUtils
 		public IEnumerable<string> GetAllVariantLanguagesFound(bool isTargetNeeded)
 		{
 			if(isTargetNeeded)
-				return File.Body.TransUnits.Select(tu => tu.Source).Select(v => v.Lang).Union(File.Body.TransUnits.Select(tu => tu.Target).Where(x => x.Lang != null).Select(v => v.Lang)).Distinct();
+				return File.Body.TransUnits.Select(tu => tu.Source).Where(s => (s != null && s.Lang != null)).Select(s => s.Lang)
+					.Union(File.Body.TransUnits.Select(tu => tu.Target).Where(t => (t != null && t.Lang != null)).Select(t => t.Lang)).Distinct();
 			else
-				return File.Body.TransUnits.Select(tu => tu.Source).Select(v => v.Lang);
+				return File.Body.TransUnits.Select(tu => tu.Source).Where(s => (s != null && s.Lang != null)).Select(s => s.Lang).Distinct();
 		}
 
 		/// ------------------------------------------------------------------------------------
@@ -150,7 +151,6 @@ namespace L10NSharp.XLiffUtils
 		/// This helps with the common case were we just changed the hierarchical organizaiton of the id,
 		/// that is, the parts of the id before th final '.'.
 		/// </summary>
-		/// <param name="id"></param>
 		 public TransUnit GetTransUnitForOrphan(TransUnit orphan)
 		{
 			 return File.Body.GetTransUnitForOrphan(orphan);
--- a/src/L10NSharp/XLiffUtils/XLiffFile.cs
+++ b/src/L10NSharp/XLiffUtils/XLiffFile.cs
@@ -33,11 +33,9 @@ namespace L10NSharp.XLiffUtils
 			DataType = "plaintext";
 		}
 
-		private int _transUnitId;
-		private bool _idsVerified;
-		protected XLiffHeader _header = new XLiffHeader();
+		protected XLiffHeader _header/* = new XLiffHeader()*/;
 		protected XLiffBody _body = new XLiffBody();
-		private List<TransUnit> _transUnits = new List<TransUnit>();
+
 		#region Properties
 
 		/// ------------------------------------------------------------------------------------
@@ -72,6 +70,15 @@ namespace L10NSharp.XLiffUtils
 		[XmlAttribute("source-language")]
 		public string SourceLang { get; set; }
 
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Gets or sets the target language found in the XLiff file.
+		/// </summary>
+		/// ------------------------------------------------------------------------------------
+		[XmlAttribute("target-language")]
+		public string TargetLang { get; set; }
+
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
 		/// Gets or sets the product version found in the XLiff file.
@@ -96,6 +103,15 @@ namespace L10NSharp.XLiffUtils
 		[XmlAttribute("datatype")]
 		public string DataType { get; set; }
 
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Gets or sets the hard linebreak replacement string.  This is the literal value displayed
+		/// to the translator (in the L10nSharp GUI) to indicate a hard linebreak in the source text.  It
+		/// defaults to \n.
+		/// </summary>
+		/// ------------------------------------------------------------------------------------
+		[XmlAttribute("hard-linebreak-replacement", Namespace=XLiffXmlSerializationHelper.kSilNamespace)]
+		public string HardLineBreakReplacement { get; set; }
 		#endregion
     }
 
--- a/src/L10NSharp/XLiffUtils/XLiffHeader.cs
+++ b/src/L10NSharp/XLiffUtils/XLiffHeader.cs
@@ -10,6 +10,7 @@
 // </remarks>
 // ---------------------------------------------------------------------------------------------
 using System.Xml.Serialization;
+using System.Collections.Generic;
 
 namespace L10NSharp.XLiffUtils
 {
@@ -25,18 +26,14 @@ namespace L10NSharp.XLiffUtils
 		/// ------------------------------------------------------------------------------------
 		public XLiffHeader()
 		{
-
+			Notes = new List<XLiffNote>();
         }
 
 		#region Properties
-		private XLiffNote _note = new XLiffNote();
 
 		[XmlElement("note")]
-		public XLiffNote Note
-		{
-			get { return _note; }
-			set { _note = value; }
-		}
+		public List<XLiffNote> Notes { get; private set; }
+
 		#endregion
 	}
 
--- a/src/L10NSharp/XLiffUtils/XLiffNote.cs
+++ b/src/L10NSharp/XLiffUtils/XLiffNote.cs
@@ -26,11 +26,37 @@ namespace L10NSharp.XLiffUtils
 		#region Properties
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
-		/// Gets or sets the type.
+		/// Gets or sets the language of the note.  Optional.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
-		[XmlAttribute("type")]
-		public string Type { get; set; }
+		[XmlAttribute("xml:lang")]
+		public string NoteLang { get; set; }
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Gets or sets who wrote the note.  Optional
+		/// </summary>
+		/// ------------------------------------------------------------------------------------
+		[XmlAttribute("from")]
+		public string From { get; set; }
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Gets or sets the priority of the note.  Optional.  If set, 1=high, 10=low.
+		/// </summary>
+		/// ------------------------------------------------------------------------------------
+		[XmlAttribute("priority"), System.ComponentModel.DefaultValue(0)]
+		public int Priority { get; set; }
+
+		/// ------------------------------------------------------------------------------------
+		/// <summary>
+		/// Gets or sets which element the note applies to, the source or the target.  Optional.
+		/// </summary>
+		/// ------------------------------------------------------------------------------------
+		[XmlAttribute("annotates")]
+		public string Annotates { get; set; }
+
+
 
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
@@ -39,18 +65,6 @@ namespace L10NSharp.XLiffUtils
 		/// ------------------------------------------------------------------------------------
 		[XmlText]
 		public string Text { get; set; }
-
-        /// ------------------------------------------------------------------------------------
-        /// <summary>
-		/// Gets the list of notes in the document header.
-        /// </summary>
-        /// ------------------------------------------------------------------------------------
-		[XmlElement("note")]
-		public List<XLiffNote> Notes
-        {
-            get { return _notes; }
-            set { _notes = value; }
-        }
         #endregion
 
         #region Methods
@@ -59,7 +73,7 @@ namespace L10NSharp.XLiffUtils
         {
             return new XLiffNote
             {
-                Type = Type,
+                NoteLang = NoteLang,
                 Text = Text
             };
         }
@@ -71,7 +85,7 @@ namespace L10NSharp.XLiffUtils
         /// ------------------------------------------------------------------------------------
         public bool IsEmpty
         {
-            get { return (string.IsNullOrEmpty(Type) && string.IsNullOrEmpty(Text)); }
+			get { return (string.IsNullOrEmpty(Text)); }
         }
 
         /// ------------------------------------------------------------------------------------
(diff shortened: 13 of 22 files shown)


JohnThomson:
Review status: 0 of 21 files reviewed at latest revision, 15 unresolved discussions.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 246 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowDFks8xP69hwV6aHv:-KowDFks8xP69hwV6aHw:b376hm0) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/LocalizedStringCache.cs#L246)):*
> ```Smalltalk
> 			xliffDoc.File.HardLineBreakReplacement = s_literalNewline;
> 			if (OwningManager != null && OwningManager.Name != null)
> 				xliffDoc.File.Original = OwningManager.Name + ".dll";
> ```

Not sure how much control we have here, but it seems to me that an xliffDoc's unqualified File ought to be the XML file it lives in; we need a more suitable name for the DLL it provides translations for.

---

*[src/L10NSharp/TransUnitUpdater.cs, line 132 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowDkTNC29OCmu68F8v:-KowDkTNC29OCmu68F8w:b-3wlich) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/TransUnitUpdater.cs#L132)):*
> ```Smalltalk
> 					_updated = true;
> 					tu.RemoveVariant(tuv);
> 					if ((tu.Source == null || tu.Source.Value == null) &&
> ```

I think you can now use tu.Source?.Value == null

---

*[src/L10NSharp/CodeReader/StringExtractor.cs, line 89 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowE1zDAUSyepMfBjSw:-KowE1zDAUSyepMfBjSx:b-rsdtnr) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/CodeReader/StringExtractor.cs#L89)):*
> ```Smalltalk
> 					//Debug.WriteLine(String.Format("DEBUG: StringExtractor.DoExtractingWork() loaded resources for {0}", type.FullName));
> 				}
> 				#pragma warning disable CS0168
> ```

May be worth commenting what warning you are disabling?

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 45 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowEDW4ZuCzzGxeF99D:-KowEDW4ZuCzzGxeF99E:b-qujvh0) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/TransUnit.cs#L45)):*
> ```Smalltalk
> 		/// <summary>
> 		/// Gets the type of translation unit.
> ```

Comment wants update

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 53 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowEIpXP6flW4dmLSXk:-KowEIpXP6flW4dmLSXl:b-4rnpt) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/TransUnit.cs#L53)):*
> ```Smalltalk
> 		/// <summary>
> 		/// Gets the source of the translation unit.
> ```

Saying this about a property called Source in a class called TransUnit doesn't add much. If it deserves a comment at all, it should indicate what sort of source (a file path? Name of the human translator? Name of a language? Text of what was translated?)

---

*[src/L10NSharp/XLiffUtils/XLiffDocument.cs, line 89 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowIm5zeUm6sXSX6fLX:-KowIm5zeUm6sXSX6fLY:b-b4oief) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffDocument.cs#L89)):*
> ```Smalltalk
> 			if(isTargetNeeded)
> 				return File.Body.TransUnits.Select(tu => tu.Source).Where(s => (s != null && s.Lang != null)).Select(s => s.Lang)
> 					.Union(File.Body.TransUnits.Select(tu => tu.Target).Where(t => (t != null && t.Lang != null)).Select(t => t.Lang)).Distinct();
> ```

This logic doesn't seem to correspond to "isTargetNeeded". I expected this version to select the languages where language and target are BOTH not-null, not where EITHER is not null. Maybe 'targetWanted' to suggest that we want targets as well as sources?
At a higher level, I wonder why this is useful. Seems we're mainly interested in the target languages, and maybe sometimes might also care about the source...if both of those mean what I think (i.e., source is just the English, and target is the actual localization data this xliff is providing).
Then there's the issue you discovered yesterday, where code is trying to enumerate languages in one xliff where it should be enumerating languages each represented by a separate xliff...this method may go away altogether?
Feel free to just ignore improving this for this PR if it's all going away in the next :-)

---

*[src/L10NSharp/XLiffUtils/XLiffFile.cs, line 76 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowJGWHUMZJE8MMWVUe:-KowJGWHUMZJE8MMWVUf:bd70zcq) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffFile.cs#L76)):*
> ```Smalltalk
> 		/// <summary>
> 		/// Gets or sets the source language found in the XLiff file.
> ```

we seem to be contrasting source and target, so why is the source language property called TargetLang?

---

*[src/L10NSharp/XLiffUtils/XLiffFile.cs, line 108 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowJS3F4YX9oXxXySgw:-KowJS3F4YX9oXxXySgx:b7oyy1n) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffFile.cs#L108)):*
> ```Smalltalk
> 		/// <summary>
> 		/// Gets or sets the hard linebreak replacement string.
> ```

Is there something more useful to say? For example, is this something that will occur in the output when there's a real hard line break in the file, or something that occurs in the file and will be changed to a hard line break in the strings returned to the client?

---

*[src/L10NSharp/XLiffUtils/XLiffNote.cs, line 88 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowJz9uVJvv5zHjee1h:-KowJz9uVJvv5zHjee1i:b-9f533h) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffNote.cs#L88)):*
> ```Smalltalk
>         public bool IsEmpty
>         {
>             get { return (string.IsNullOrEmpty(NoteLang) && string.IsNullOrEmpty(Text)); }
> ```

I find it surprising that a Note with no text is non-empty, even if it has a language. (I could imagine that is it empty, or at least useless, if EITHER is missing.)

---

*[src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs, line 117 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowKQhg-8d5kgPNgIER:-KowKQhg-8d5kgPNgIES:bfvgcws) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs#L117)):*
> ```Smalltalk
> 				using (StringWriter writer = new StringWriter(output))
> 				{
> 					XmlSerializerNamespaces nameSpace = new XmlSerializerNamespaces();
> ```

Should the name be plural?

---

*[src/L10NSharpTests/XLiffUtilsTests.cs, line 27 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowUDo2IZxOqpLxWlUd:-KowUDo2IZxOqpLxWlUe:b-t3v766) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharpTests/XLiffUtilsTests.cs#L27)):*
> ```Smalltalk
> 
> 		[Test]
> 		public void TestGetAllVariantLanguagesFound()
> ```

Should this be enhanced to verify that it can find more than one of each? (Assuming we still need this method at all once you sort out the deeper logic errors :-)

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 14 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowUi8onY_m6LYPddKW:-KowUi8onY_m6LYPddKX:bcrgku8) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L14)):*
> ```XSLT
> 	
> 	<xsl:variable name="srclang"><xsl:value-of select="/tmx/header/@adminlang"/></xsl:variable>
> 	<xsl:variable name="targlang"><xsl:value-of select="/tmx/header/@srclang"/></xsl:variable>
> ```

Aaargh! So what tmx calls 'source' is what xliff calls 'target', and xliff calls something else (tmx 'admin') source? No wonder comments are getting confused! Not sure what you can do, but try to be sure that every place a 'Source' property is declared it's clear whether we mean the tmx or xliff sense.

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 30 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowVRe-drbCWUj4nOCv:-KowVRe-drbCWUj4nOCw:b-k4imbv) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L30)):*
> ```XSLT
> 	<xsl:template match="tmx">
> 		<xsl:element name="file">
> 			<xsl:attribute name="original">CHANGE-ME.dll</xsl:attribute>
> ```

Did you mean to change this before merging, or is code going to change this later?

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 34 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowVffpWM2mtlN_-VJc:-KowVffpWM2mtlN_-VJd:b-i1yj6a) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L34)):*
> ```XSLT
> 			<xsl:if test="$srclang != $targlang">
> 				<xsl:attribute name="target-language"><xsl:value-of select="$targlang"/></xsl:attribute>
> 			</xsl:if>
> ```

These lines seem inconsistent with how you got values for srclang and targlang above.

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 84 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowWzZk-EZ0Gcho2q8E:-KowWzZk-EZ0Gcho2q8F:bh1zces) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L84)):*
> ```XSLT
> 		<xsl:if test="not(following-sibling::tuv)">
> 			<xsl:apply-templates select="preceding-sibling::note"/>
> 		</xsl:if>
> ```

Would it be simpler to just do xsl:apply-templates select ="note" following the tuv select at the parent level?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16)*

StephenMcConnel:
Review status: 0 of 21 files reviewed at latest revision, 15 unresolved discussions.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 246 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowDFks8xP69hwV6aHv:-Kow_j9XS0tVoxMYarx4:bz8cfsp) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/LocalizedStringCache.cs#L246)):*

Not sure how much control we have here, but it seems to me that an xliffDoc's unqualified File ought to be the XML file it lives in; we need a more suitable name for the DLL it provides translations for.
</blockquote></details>

Quoting from the xliff 1.2 standard, "The original attribute specifies the name of the original file from which the contents of a <file> element has been extracted."
This value doesn't really exist in our setup since we're extracting strings from a large number of C# (or Typescript) files or assemblies.  We could complicate things by having multiple file elements in the xliff, one per actual source file with original pointing to that exact source file.  However, some strings are used in multiple places in the program with identical id values, so that breaks down also.  Changing from ".dll" to ".xlf" doesn't really fit the intention of the standard any better than what is being done currently, and will likely look like a circular reference to the current xliff file itself.
(NB, the xliff standard does allow multiple file elements inside the xliff element, although our implementation has only one.)

Summary:  I'm not inclined to change this at the moment.  We could discuss whether to implement multiple file elements with more meaningful values for the "original" attribute, but I'm not eager to go there if we don't have to.

---

*[src/L10NSharp/TransUnitUpdater.cs, line 132 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowDkTNC29OCmu68F8v:-KowbSrXWEoEk1qI76WT:b-bmpj30) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/TransUnitUpdater.cs#L132)):*

I think you can now use tu.Source?.Value == null
</blockquote></details>

That's nice to know, but I'm not convinced it's as clear for a casual reader of the code.  At least not until the syntax has been used a lot longer.  And there's hundreds of places in our code that could changed in this way...

---

*[src/L10NSharp/CodeReader/StringExtractor.cs, line 89 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowE1zDAUSyepMfBjSw:-KowdSomKZINz2HhdEp2:befsf8t) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/CodeReader/StringExtractor.cs#L89)):*

May be worth commenting what warning you are disabling?
</blockquote></details>

Done.  for all of the #pragmas that I added.

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 45 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowEDW4ZuCzzGxeF99D:-KoweU9c2_F9nURSVVk0:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/TransUnit.cs#L45)):*

Comment wants update
</blockquote></details>

Done.

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 53 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowEIpXP6flW4dmLSXk:-Kowf57Bz1RBHNzeHnJj:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/TransUnit.cs#L53)):*

Saying this about a property called Source in a class called TransUnit doesn't add much. If it deserves a comment at all, it should indicate what sort of source (a file path? Name of the human translator? Name of a language? Text of what was translated?)
</blockquote></details>

Done.

---

*[src/L10NSharp/XLiffUtils/XLiffDocument.cs, line 89 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowIm5zeUm6sXSX6fLX:-KowjgmjlclXRQ7AmRzd:bj33lnb) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffDocument.cs#L89)):*

This logic doesn't seem to correspond to "isTargetNeeded". I expected this version to select the languages where language and target are BOTH not-null, not where EITHER is not null. Maybe 'targetWanted' to suggest that we want targets as well as sources?
At a higher level, I wonder why this is useful. Seems we're mainly interested in the target languages, and maybe sometimes might also care about the source...if both of those mean what I think (i.e., source is just the English, and target is the actual localization data this xliff is providing).
Then there's the issue you discovered yesterday, where code is trying to enumerate languages in one xliff where it should be enumerating languages each represented by a separate xliff...this method may go away altogether?
Feel free to just ignore improving this for this PR if it's all going away in the next :-)
</blockquote></details>

Although I got this method working more or less without crashing in Bloom, I'm not at all convinced it serves a useful purpose.
The isTargetNeeded flag indicates that both source and target elements are to be checked.  It makes more sense when multiple target files have been merged and some have actually contributed translation units.  But I'm not sure that the merging operation ever makes sense.
I'd rather leave any changes to this to the next PR since it might well be removed.

---

*[src/L10NSharp/XLiffUtils/XLiffFile.cs, line 76 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowJGWHUMZJE8MMWVUe:-KowfN7oToC1Q-MkYkfK:b-5kgqb1) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffFile.cs#L76)):*

we seem to be contrasting source and target, so why is the source language property called TargetLang?
</blockquote></details>

Done.  comment should have said target instead of source.

---

*[src/L10NSharp/XLiffUtils/XLiffFile.cs, line 108 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowJS3F4YX9oXxXySgw:-KowvHQOcvCfAEV2sR5Y:b-djaewc) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffFile.cs#L108)):*

Is there something more useful to say? For example, is this something that will occur in the output when there's a real hard line break in the file, or something that occurs in the file and will be changed to a hard line break in the strings returned to the client?
</blockquote></details>

I added a bit more to the comment to try to explain what this is used for.

---

*[src/L10NSharp/XLiffUtils/XLiffNote.cs, line 88 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowJz9uVJvv5zHjee1h:-Kowgitn59uQdrEnKejp:bshgaby) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffNote.cs#L88)):*

I find it surprising that a Note with no text is non-empty, even if it has a language. (I could imagine that is it empty, or at least useless, if EITHER is missing.)
</blockquote></details>

changed to check only the text.  The NoteLang value is often empty, assumed to be either English or whatever the default source language is.

---

*[src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs, line 117 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowKQhg-8d5kgPNgIER:-Kowh3hj9blsBnWDZUZp:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs#L117)):*

Should the name be plural?
</blockquote></details>

Done.

---

*[src/L10NSharpTests/XLiffUtilsTests.cs, line 27 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowUDo2IZxOqpLxWlUd:-Koww0iCBEWkYyVWGDAi:b-3mp8o9) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharpTests/XLiffUtilsTests.cs#L27)):*

Should this be enhanced to verify that it can find more than one of each? (Assuming we still need this method at all once you sort out the deeper logic errors :-)
</blockquote></details>

The final block does test for 2 languages being found.  With the method being passed false, I would hope that only 1 language would ever be found.  So I think the test is good enough for now, especially with the future of the method being somewhat in doubt.  :-)

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 14 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowUi8onY_m6LYPddKW:-KowhBu8J2Gg8B4ulSGN:b-s0luw6) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L14)):*

Aaargh! So what tmx calls 'source' is what xliff calls 'target', and xliff calls something else (tmx 'admin') source? No wonder comments are getting confused! Not sure what you can do, but try to be sure that every place a 'Source' property is declared it's clear whether we mean the tmx or xliff sense.
</blockquote></details>

I'm not sure whether what we were doing in TMX made any sense, but it seemed to be consistent in all of the files.

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 30 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowVRe-drbCWUj4nOCv:-KowhKvfPzf4QiJSwod3:bt1csqw) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L30)):*

Did you mean to change this before merging, or is code going to change this later?
</blockquote></details>

There's no way to get a reasonable value inside the file, so I made it something that would attract attention.  :-)  I think it will have to be manually edited in the xliff files created from tmx files.  There isn't any value in the tmx files that seems suitable to put in this attribute.  And the name of the input tmx file isn't available inside the xslt processing as far as I could determine.  Which might make a plausible (even if technically wrong) value.

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 34 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowVffpWM2mtlN_-VJc:-KowhdLUD7a4kPZH9lX9:b-9inx3k) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L34)):*

These lines seem inconsistent with how you got values for srclang and targlang above.
</blockquote></details>

why?  what xliff calls the source-language appears to marked adminlang in our tmx files, and the target-language in xliff terms is marked srclang in our tmx files.
In passing, the xml:lang attributes of all source elements is optional because it's supposed to match the file's source-language attribute in xliff.  Similarly, the xml:lang attribute of all target elements is optional and supposed to match the file's target-language attribute (especially if you follow the recommendation to have only bilingual xliff files).

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 84 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowWzZk-EZ0Gcho2q8E:-Kowhju4zOlRyTpqzx_E:b66gt5m) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L84)):*

Would it be simpler to just do xsl:apply-templates select ="note" following the tuv select at the parent level?
</blockquote></details>

good point.  :-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16)*

JohnThomson:
Looks much better. Few more minor suggestions.

---

Review status: 0 of 21 files reviewed at latest revision, 11 unresolved discussions.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 246 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowDFks8xP69hwV6aHv:-Kp0HRmY9sA-r_t1T5Ef:bea95il) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/LocalizedStringCache.cs#L246)):*

Quoting from the xliff 1.2 standard, "The original attribute specifies the name of the original file from which the contents of a <file> element has been extracted."
This value doesn't really exist in our setup since we're extracting strings from a large number of C# (or Typescript) files or assemblies.  We could complicate things by having multiple file elements in the xliff, one per actual source file with original pointing to that exact source file.  However, some strings are used in multiple places in the program with identical id values, so that breaks down also.  Changing from ".dll" to ".xlf" doesn't really fit the intention of the standard any better than what is being done currently, and will likely look like a circular reference to the current xliff file itself.
(NB, the xliff standard does allow multiple file elements inside the xliff element, although our implementation has only one.)

Summary:  I'm not inclined to change this at the moment.  We could discuss whether to implement multiple file elements with more meaningful values for the "original" attribute, but I'm not eager to go there if we don't have to.
</blockquote></details>

Not being sure what this value was even supposed to mean, I was not sure what if anything should be changed. Now that I know, I think using the name of the DLL for which it holds translations is a reasonable VALUE.  I would not have called such a thing File.Original, but it sounds as though that name is part of the standard, so I agree, let's keep it.

---

*[src/L10NSharp/UI/CustomDropDownComboBox.cs, line 16 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0IPZ-8H8i1RKb67Qn:-Kp0IPZ-8H8i1RKb67Qo:bdm9a53) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/UI/CustomDropDownComboBox.cs#L16)):*
> ```Smalltalk
> 
> 		#pragma warning disable CS0649	// field declaration is never assigned a value
> ```

So why have it at all?

---

*[src/L10NSharp/UI/L10NExtender.cs, line 37 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0IT-KTws7w8nvalGJ:-Kp0IT-KTws7w8nvalGK:b-fihtsw) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/UI/L10NExtender.cs#L37)):*
> ```Smalltalk
> 
> 		#pragma warning disable CS0414	// field is assigned a value but its value is never used
> 		private Container components = null;
> ```

Likewise

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 45 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0Ia0w53e87kDgv9-z:-Kp0Ia0w53e87kDgv90-:bffekpu) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/XLiffUtils/TransUnit.cs#L45)):*
> ```Smalltalk
> 		/// <summary>
> 		/// Gets whether the unit is "dynamic" (discovered dynamically).
> ```

Better, but "discovered dynamically" still assumes a lot of knowledge. I always have trouble remembering whether dynamic are the ones discovered by scanning the files or (I think) the ones that get missed by that and only found while running the program. Maybe that's explained in depth somewhere else...could be cross-ref? Or just leave it...maybe I'm overdoing things.

---

*[src/L10NSharp/XLiffUtils/XLiffDocument.cs, line 89 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowIm5zeUm6sXSX6fLX:-Kp0JYJ0CqDxvRuR_buB:b-gxztsp) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffDocument.cs#L89)):*

Although I got this method working more or less without crashing in Bloom, I'm not at all convinced it serves a useful purpose.
The isTargetNeeded flag indicates that both source and target elements are to be checked.  It makes more sense when multiple target files have been merged and some have actually contributed translation units.  But I'm not sure that the merging operation ever makes sense.
I'd rather leave any changes to this to the next PR since it might well be removed.
</blockquote></details>

Fair enough.

---

*[src/L10NSharp/XLiffUtils/XLiffFile.cs, line 108 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowJS3F4YX9oXxXySgw:-Kp0JlFX-EJogqCspc1Q:buw5mfi) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/L10NSharp/XLiffUtils/XLiffFile.cs#L108)):*

I added a bit more to the comment to try to explain what this is used for.
</blockquote></details>

Very helpful. That meaning had not even occurred to me as a possibility!

---

*[src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs, line 127 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0KTIBe31jpy6KIChT:-Kp0KTIBe31jpy6KIChU:b-dng3q) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs#L127)):*
> ```Smalltalk
> 				return (output.Length == 0 ? null : output.ToString());
> 			}
> 			#pragma warning disable CS0168, CS0162	// variable is declared but not used, unreachable code detected 
> ```

As things stand, we could remove the whole try/catch wrapper, unless possibly we often want to put a breakpoint on the throw.

---

*[src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs, line 158 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0KgvdMjOzhBPQjsBS:-Kp0KgvdMjOzhBPQjsBT:b-ytwyha) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs#L158)):*
> ```Smalltalk
> 				}
> 			}
> 			#pragma warning disable CS0168, CS0162	// variable is declared but not used, unreachable code detected
> ```

and again

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 14 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowUi8onY_m6LYPddKW:-Kp0O59OUUaV02klyxg8:b-hieuft) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L14)):*

I'm not sure whether what we were doing in TMX made any sense, but it seemed to be consistent in all of the files.
</blockquote></details>

I'm just concerned about how confusing it will be if all our method and property names are written as if 'source' means the translation, once xliff becomes the main file format and in it 'source' is the language translated FROM (which actually makes better sense to me). Seems to call for at least comments where we declare properties called Source and especially where we load a target attribute into a property called source. But the main thing is that in all our C# etc Source consistently means the translation.

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 30 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowVRe-drbCWUj4nOCv:-Kp0MZvCoXxDmGOJdhXt:bmlbgsr) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L30)):*

There's no way to get a reasonable value inside the file, so I made it something that would attract attention.  :-)  I think it will have to be manually edited in the xliff files created from tmx files.  There isn't any value in the tmx files that seems suitable to put in this attribute.  And the name of the input tmx file isn't available inside the xslt processing as far as I could determine.  Which might make a plausible (even if technically wrong) value.
</blockquote></details>

xslt 2 has functions base-uri and document-uri which I think should retrieve the tmx file name if applied to a node from that file. Try <xsl:variable name="document-uri" select="document-uri(.)"/>. May need to strip off unwanted path elements.

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 34 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowVffpWM2mtlN_-VJc:-Kp0LZ2gL4yGcTRiaby2:b2qr03m) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L34)):*

why?  what xliff calls the source-language appears to marked adminlang in our tmx files, and the target-language in xliff terms is marked srclang in our tmx files.
In passing, the xml:lang attributes of all source elements is optional because it's supposed to match the file's source-language attribute in xliff.  Similarly, the xml:lang attribute of all target elements is optional and supposed to match the file's target-language attribute (especially if you follow the recommendation to have only bilingual xliff files).
</blockquote></details>

You're right, not sure what I was thinking.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-:-Kp0Nj7jFiSlTni0ZEBS:b6ckqyx)*

JohnThomson:
Reviewed 12 of 21 files at r1, 9 of 9 files at r2.
Review status: all files reviewed at latest revision, 11 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16)*

StephenMcConnel:
Review status: 15 of 22 files reviewed at latest revision, 11 unresolved discussions.

---

*[src/L10NSharp/UI/CustomDropDownComboBox.cs, line 16 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0IPZ-8H8i1RKb67Qn:-Kp1mLsoJ-rqtoXtIrI9:b-ofpbys) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/UI/CustomDropDownComboBox.cs#L16)):*

So why have it at all?
</blockquote></details>

Okay, I've removed this even though it had a ripple effect that removed two methods and simplified a property definition greatly.  I've removed all the item requiring a #pragma that appeared to be reasonable to remove.

---

*[src/L10NSharp/UI/L10NExtender.cs, line 37 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0IT-KTws7w8nvalGJ:-Kp1mYg-iNpmcwx_0K7u:b-jrzv9z) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/UI/L10NExtender.cs#L37)):*

Likewise
</blockquote></details>

This one is needed for Designer support apparently.  I copied the comment from where it's given a value.

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 45 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0Ia0w53e87kDgv9-z:-Kp1oP-venLVIwk9TSm_:b-7w6ql3) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/XLiffUtils/TransUnit.cs#L45)):*

Better, but "discovered dynamically" still assumes a lot of knowledge. I always have trouble remembering whether dynamic are the ones discovered by scanning the files or (I think) the ones that get missed by that and only found while running the program. Maybe that's explained in depth somewhere else...could be cross-ref? Or just leave it...maybe I'm overdoing things.
</blockquote></details>

added "while the program is running" to the comment

---

*[src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs, line 127 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0KTIBe31jpy6KIChT:-Kp1osSxAJZiR7C9Sni6:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs#L127)):*

As things stand, we could remove the whole try/catch wrapper, unless possibly we often want to put a breakpoint on the throw.
</blockquote></details>

Done.

---

*[src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs, line 158 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-Kp0KgvdMjOzhBPQjsBS:-Kp1pnNYAUfyX1MLVMkO:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d1535b827f36febc651ea83468d4a2f5d41a3730/src/L10NSharp/XLiffUtils/XLiffXmlSerializationHelper.cs#L158)):*

and again
</blockquote></details>

Done.

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 30 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowVRe-drbCWUj4nOCv:-Kp1sk5G4XxDb2B-mqNK:b-gmpt3x) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L30)):*

xslt 2 has functions base-uri and document-uri which I think should retrieve the tmx file name if applied to a node from that file. Try <xsl:variable name="document-uri" select="document-uri(.)"/>. May need to strip off unwanted path elements.
</blockquote></details>

Unfortunately, the xslt processor I'm using only handles version 1.0 XSL.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16)*

StephenMcConnel:
Review status: 15 of 22 files reviewed at latest revision, 11 unresolved discussions.

---

*[src/xslt/l10nSharpTmxToXliff1-2.xsl, line 14 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-KowUi8onY_m6LYPddKW:-Kp1yCOR3cEeA1wKxRTx:b-ujqlae) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a14074ff5f3831835b6f9932f77049b67b25ba42/src/xslt/l10nSharpTmxToXliff1-2.xsl#L14)):*

I'm just concerned about how confusing it will be if all our method and property names are written as if 'source' means the translation, once xliff becomes the main file format and in it 'source' is the language translated FROM (which actually makes better sense to me). Seems to call for at least comments where we declare properties called Source and especially where we load a target attribute into a property called source. But the main thing is that in all our C# etc Source consistently means the translation.
</blockquote></details>

I'm not sure I want to fully address this concern in this PR.  But it is something to think about as I continue on work on the code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16)*

JohnThomson:
:lgtm:

---

Reviewed 7 of 7 files at r3.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/16#-:-KpFYgYng8scPuX04ZEZ:bnfp4nl)*

JohnThomson:
I don't seem to have permission to merge this. @hatton?

hatton:
@JohnThomson please try now.

PR #17, opened 2017-07-19
Allow folder structure for xliff files

StephenMcConnel:
Also finish getting new xliff based code to work with Bloom.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/17)

### DIFF
--- a/src/L10NSharp/LocalizationManager.cs
+++ b/src/L10NSharp/LocalizationManager.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.Diagnostics;
 using System.Drawing;
 using System.Globalization;
 using System.IO;
@@ -29,6 +30,15 @@ namespace L10NSharp
 		/// </summary>
 		public static event EventHandler LaunchingLocalizationDialog;
 		public static event EventHandler ClosingLocalizationDialog;
+
+		/// <summary>
+		/// Flag that the program organizes translation files by folder rather than by filename.
+		/// That is, localization/en/AppName.xlf (English) and localization/id/AppName.xlf (Indonesian)
+		/// instead of localization/AppName.en.xlf and localization/AppName.id.xlf.
+		/// Note that this must be set before creating any LocalizationManager objects.
+		/// The default is the old way of organizing (by filename).
+		/// </summary>
+		public static bool UseLanguageCodeFolders;
 
 		private static string s_uiLangId;
 		private static List<string> s_fallbackLanguageIds = new List<string>(new[] { kDefaultLang });
@@ -260,19 +270,16 @@ namespace L10NSharp
 			var fileStream = File.Open(DefaultStringFilePath, FileMode.Create, FileAccess.ReadWrite, FileShare.None);
 			fileStream.Close();
 
-			var XliffDoc = LocalizedStringCache.CreateEmptyStringFile();
-			XliffDoc.File.ProductVersion = AppVersion;
-			XliffDoc.File.Original = Id + ".dll";
-			var tuUpdater = new TransUnitUpdater(XliffDoc);
+			var stringCache = new LocalizedStringCache(this, false);
 
 			using (var dlg = new InitializationProgressDlg(Name, _applicationIcon, namespaceBeginnings))
 			{
 				dlg.ShowDialog();
 				foreach (var locInfo in dlg.ExtractedInfo)
-					tuUpdater.Update(locInfo);
+					stringCache.UpdateLocalizedInfo(locInfo);
 			}
 
-			XliffDoc.Save(DefaultStringFilePath);
+			stringCache.SaveIfDirty(stringCache.XliffDocuments.Keys);
 		}
 
 		/// <summary> Sometimes, on Linux, there is an empty DefaultStringFile.  This causes problems. </summary>
@@ -346,7 +353,7 @@ namespace L10NSharp
 			var allLangs = groups.Select(g => g.First()).ToList();
 
 			var langsHavinglocalizations = (LoadedManagers == null ? new List<string>() :
-				LoadedManagers.Values.SelectMany(lm => lm.StringCache.XliffDocument.GetAllVariantLanguagesFound(false))
+				LoadedManagers.Values.SelectMany(lm => lm.GetAvailableUILanguageTags())
 				.Distinct().ToList());
 
 			// BL-1011: Make sure cultures that have existing localizations are included
@@ -378,21 +385,27 @@ namespace L10NSharp
 		}
 
 		/// <summary>
-		/// Return the language tags for those languages that have been localized for the given program
-		/// in its localization folder.
+		/// Return the language tags for those languages that have been localized for the given program.
 		/// </summary>
-		public static IEnumerable<string> GetAvailableUILanguageTags(string localizationFolder, string programName)
+		public IEnumerable<string> GetAvailableUILanguageTags()
 		{
-			var tags = new List<string>();
-			if (!Directory.Exists(localizationFolder))
-				return tags;
-			foreach (var filepath in Directory.GetFiles(localizationFolder, programName + ".*" + kFileExtension))
+			return StringCache.XliffDocuments.Keys.ToList();
+		}
+
+		/// <summary>
+		/// If the given file exists, return its parent folder name as a language tag if it
+		/// appears to be valid (2 characters long or "zh-CN").  Otherwise return null.
+		/// </summary>
+		private static string GetLanguageTagFromFilePath(string xliffFile)
+		{
+			Debug.Assert(UseLanguageCodeFolders);
+			if (File.Exists(xliffFile))
 			{
-				var filename = Path.GetFileNameWithoutExtension(filepath);
-				var tag = filename.Substring(programName.Length + 1);
-				tags.Add(tag);
+				var langId = Path.GetFileName(Path.GetDirectoryName(xliffFile));
+				if ((langId.Length == 2 || langId == "zh-CN"))
+					return langId;
 			}
-			return tags;
+			return null;
 		}
 		#endregion
 
@@ -536,23 +549,57 @@ namespace L10NSharp
 				HashSet<string> langIdsOfCustomizedLocales = new HashSet<string>();
 				string langId;
 				if (_customXliffFileFolder != null && Directory.Exists(_customXliffFileFolder))
-					foreach (var XliffFile in Directory.GetFiles(_customXliffFileFolder, Id + ".*" + kFileExtension))
+				{
+					if (UseLanguageCodeFolders)
 					{
-						langId = GetLangIdFromXliffFileName(XliffFile);
-						if (langId != kDefaultLang) // should never happen for customized languages
+						foreach (var folder in Directory.GetDirectories(_customXliffFileFolder))
 						{
-							langIdsOfCustomizedLocales.Add(langId);
-							yield return XliffFile;
+							var XliffFile = Path.Combine(folder, Id + kFileExtension);
+							langId = GetLanguageTagFromFilePath(XliffFile);
+							if (!String.IsNullOrEmpty(langId) && langId != kDefaultLang)
+							{
+								langIdsOfCustomizedLocales.Add(langId);
+								yield return XliffFile;
+							}
 						}
 					}
+					else
+					{
+						foreach (var XliffFile in Directory.GetFiles(_customXliffFileFolder, Id + ".*" + kFileExtension))
+						{
+							langId = GetLangIdFromXliffFileName(XliffFile);
+							if (langId != kDefaultLang)
+							{
+								langIdsOfCustomizedLocales.Add(langId);
+								yield return XliffFile;
+							}
+						}
+					}
+				}
 				if (_installedXliffFileFolder != null)
 				{
-					foreach (var XliffFile in Directory.GetFiles(_installedXliffFileFolder, Id + ".*" + kFileExtension))
+					if (UseLanguageCodeFolders)
 					{
-						langId = GetLangIdFromXliffFileName(XliffFile);
-						if (langId != kDefaultLang &&    //Don't return the english Xliff here because we separately process it first.
-							!langIdsOfCustomizedLocales.Contains(langId))
-							yield return XliffFile;
+						foreach (var folder in Directory.GetDirectories(_installedXliffFileFolder))
+						{
+							var XliffFile = Path.Combine(folder, Id + kFileExtension);
+							langId = GetLanguageTagFromFilePath(XliffFile);
+							if (!String.IsNullOrEmpty(langId) && langId != kDefaultLang)
+							{
+								langIdsOfCustomizedLocales.Add(langId);
+								yield return XliffFile;
+							}
+						}
+					}
+					else
+					{
+						foreach (var XliffFile in Directory.GetFiles(_installedXliffFileFolder, Id + ".*" + kFileExtension))
+						{
+							langId = GetLangIdFromXliffFileName(XliffFile);
+							if (langId != kDefaultLang &&    //Don't return the english Xliff here because we separately process it first.
+								!langIdsOfCustomizedLocales.Contains(langId))
+								yield return XliffFile;
+						}
 					}
 				}
 			}
@@ -710,6 +757,7 @@ namespace L10NSharp
 		/// ------------------------------------------------------------------------------------
 		private string GetLangIdFromXliffFileName(string fileName)
 		{
+			Debug.Assert(!UseLanguageCodeFolders);
 			fileName = fileName.Substring(0, fileName.Length - kFileExtension.Length);
 			int i = fileName.LastIndexOf('.');
 			return (i < 0 ? null : fileName.Substring(i + 1));
@@ -1097,7 +1145,14 @@ namespace L10NSharp
 		/// ------------------------------------------------------------------------------------
 		public static string GetXliffFileNameForLanguage(string appId, string langId)
 		{
-			return string.Format("{0}.{1}" + kFileExtension, appId, langId);
+			if (UseLanguageCodeFolders)
+			{
+				return Path.Combine(langId, appId + kFileExtension);
+			}
+			else
+			{
+				return string.Format("{0}.{1}{2}", appId, langId, kFileExtension);
+			}
 		}
 
 		/// ------------------------------------------------------------------------------------
--- a/src/L10NSharp/LocalizedStringCache.cs
+++ b/src/L10NSharp/LocalizedStringCache.cs
@@ -1,7 +1,8 @@
 ﻿using System;
-using System.Linq;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Security;
 using System.Text;
 using System.Windows.Forms;
@@ -13,10 +14,6 @@ namespace L10NSharp
 	/// ----------------------------------------------------------------------------------------
 	internal class LocalizedStringCache
 	{
-		internal const string kPriorityPropTag = "x-priority";
-		internal const string kGroupPropTag = "x-group";
-		internal const string kCategoryPropTag = "x-category";
-		internal const string kDiscoveredDyanmically = "x-dynamic";
 		internal const string kToolTipSuffix = "_ToolTip_";
 		internal const string kShortcutSuffix = "_ShortcutKeys_";
 
@@ -24,18 +21,25 @@ namespace L10NSharp
 		// character which, when included, messes up the display of text in controls.
 		internal const string kOSRealNewline = "\n";
 
-		private readonly string _ampersandReplacement = "|amp|";
+		internal const string kDefaultAmpersandReplacement = "|amp|";
+		internal static string _ampersandReplacement = kDefaultAmpersandReplacement;
 
 		// This is the symbol for a newline that users put in their localized text when
 		// they want a real newline inserted. The program will replace literal newlines
 		// with the value of kOSNewline.
-		internal static string s_literalNewline = "\\n";
+		internal const string kDefaultNewlineReplacement = "\\n";
+		internal static string s_literalNewline = kDefaultNewlineReplacement;
 
 		private readonly TransUnitUpdater _tuUpdater;
 
 		internal List<LocTreeNode> LeafNodeList { get; private set; }
 		internal LocalizationManager OwningManager { get; private set; }
-		public XLiffDocument XliffDocument { get; private set; }
+		private XLiffDocument DefaultXliffDocument { get; set; }	// matches LanguageManager.kDefaultLanguage
+
+		/// <summary>
+		/// Record the xliff document loaded for each language.
+		/// </summary>
+		internal readonly Dictionary<string, XLiffDocument> XliffDocuments = new Dictionary<string, XLiffDocument>();
 
 		#region Loading methods
 		/// ------------------------------------------------------------------------------------
@@ -43,29 +47,35 @@ namespace L10NSharp
 		/// Loads the string cache from all the specified Xliff files
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
-		internal LocalizedStringCache(LocalizationManager owningManager)
+		internal LocalizedStringCache(LocalizationManager owningManager, bool loadAvailableXliffFiles = true)
 		{
 			OwningManager = owningManager;
-			XliffDocument = CreateEmptyStringFile();
-			XliffDocument.File.Original = owningManager.Name + ".dll";
-			try
+			if (loadAvailableXliffFiles)
 			{
-				MergeXliffFilesIntoCache(OwningManager.XliffFilenamesToAddToCache);
+				try
+				{
+					MergeXliffFilesIntoCache(OwningManager.XliffFilenamesToAddToCache);
+				}
+				catch (Exception e)
+				{
+					MessageBox.Show("Error occurred reading localization file:" + Environment.NewLine + e.Message,
+						Application.ProductName, MessageBoxButtons.OK, MessageBoxIcon.Warning);
+					LocalizationManager.SetUILanguage(LocalizationManager.kDefaultLang, false);
+				}
 			}
-			catch (Exception e)
+			else
 			{
-				MessageBox.Show("Error occurred reading localization file:" + Environment.NewLine + e.Message,
-					Application.ProductName, MessageBoxButtons.OK, MessageBoxIcon.Warning);
-				LocalizationManager.SetUILanguage(LocalizationManager.kDefaultLang, false);
+				DefaultXliffDocument = CreateEmptyStringFile();
+				DefaultXliffDocument.File.Original = owningManager.Name + ".dll";
+				XliffDocuments.Add(LocalizationManager.kDefaultLang, DefaultXliffDocument);
 			}
+			_tuUpdater = new TransUnitUpdater(this);
 
-			_tuUpdater = new TransUnitUpdater(XliffDocument);
-
-			var replacement = XliffDocument.File.GetPropValue("x-ampersandreplacement");
+			var replacement = DefaultXliffDocument.File.AmpersandReplacement;
 			if (replacement != null)
 				_ampersandReplacement = replacement;
 
-			replacement = XliffDocument.File.HardLineBreakReplacement;
+			replacement = DefaultXliffDocument.File.HardLineBreakReplacement;
 			if (replacement != null)
 				s_literalNewline = replacement;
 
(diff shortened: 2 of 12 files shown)


JohnThomson:
Probably (per discussion this morning) too many comments. Partly because I don't know this code base well, so changes may mystify me that would be quite clear reading the whole thing. Don't feel you have to answer every question or do something about every comment...it's fine to ignore whatever isn't helpful.
Overall, I find the flow of information hard to figure out. There are many objects containing localization data...XLiffDocument, XLiffFile, XLiffBody, LocalizedStringCache, LocalizingInfo, at least one dictionary of dictionaries (lang id to dict of string id to TU), the actual TranslationUnit objects. It isn't always obvious why information is being transformed from one to another. If there aren't already, it would be good to have class comments that spell out what each is responsible for. But this PR is probably not the place to fix that.

---

Reviewed 12 of 12 files at r1.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[src/L10NSharp/LocalizationManager.cs, line 35 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVBk-l56E1OKI3J3V9:-KpVBk-l56E1OKI3J3VA:b1odfaw) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizationManager.cs#L35)):*
> ```Smalltalk
> 		/// <summary>
> 		/// Flag that the program organizes translation files by folder rather than by filename.
> 		/// That is, localization/AppName.xlf (English) and localization/id/AppName.xlf (Indonesian)
> ```

So English is specifically an exception? Not localization/en/AppName.xlf?

---

*[src/L10NSharp/LocalizationManager.cs, line 394 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpV9yUkHFuTRBCJfE26:-KpV9yUkHFuTRBCJfE27:b1s6kil) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizationManager.cs#L394)):*
> ```Smalltalk
> 			foreach (var key in StringCache.XliffDocuments.Keys)
> 				tags.Add(key);
> 			return tags;
> ```

return StringCache.XliffDocuments.Keys.ToList()? (Would also work without the ToList unless the evaluating is expensive or the caller is likely to modify the list.)

---

*[src/L10NSharp/LocalizationManager.cs, line 407 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVAVL2SGSWTSO6wNhT:-KpVAVL2SGSWTSO6wNhU:b-s8ek3c) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizationManager.cs#L407)):*
> ```Smalltalk
> 				var langId = Path.GetFileName(Path.GetDirectoryName(xliffFile));
> 				if ((langId.Length == 2 || langId == "zh-CN") && langId != kDefaultLang)
> 					return langId;
> ```

Should this have a different implementation if UseLanguageCodeFolders is false? (I see below that a different routine is called in that case...maybe this one could just Assert(UseLanguageCodeFolders)?
Should it have a special case for English (which apparently lives in the root folder)?

---

*[src/L10NSharp/LocalizedStringCache.cs, line 107 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVECBDV8Nw4UjymHk2:-KpVECBDV8Nw4UjymHk3:biaoc87) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L107)):*
> ```Smalltalk
> 				foreach (var tu in defaultInstalledXliffDoc.File.Body.TransUnits)
> 				{
> 					DefaultXliffDocument.File.Body.AddTransUnitOrVariantFromExisting(tu, LocalizationManager.kDefaultLang);
> ```

I wonder whether we should encapsulate the File.Body bit by giving XLiffDocument an AddTransUnitOrVariantFromExisting method? One day we might want to make the in-memory object store translations in some form more compact than an XML document and be sorry we let this implementation detail escape. (I see below that File.Body. Whatever is a common pattern, and was even before this PR. So probably not something to fix now. A thought to consider as you think about further enhancement.)

---

*[src/L10NSharp/LocalizedStringCache.cs, line 109 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVDipoIixMtA2Bb1cX:-KpVDipoIixMtA2Bb1cY:by0j4zb) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L109)):*
> ```Smalltalk
> 					DefaultXliffDocument.File.Body.AddTransUnitOrVariantFromExisting(tu, LocalizationManager.kDefaultLang);
> 					// also needed in this, to prevent things we find here from being considered orphans.
> 					defaultXliffDoc.File.Body.AddTransUnitOrVariantFromExisting(tu, LocalizationManager.kDefaultLang);
> ```

It's hard to get this straight without being able to easily navigate to declarations etc. But if defaultXliffDoc and DefaultXliffDocument are different objects, as appears to be the case, I think one of them should have a different name to help the reader figure out what is going on here.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 121 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVFL8ciHNhYTREvn0k:-KpVFL8ciHNhYTREvn0l:bdzmvkj) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L121)):*
> ```Smalltalk
> 			Exception error = null;
> 
> 			foreach (var file in xliffFiles.Where(f => Path.GetFileName(f) != OwningManager.DefaultStringFilePath))
> ```

This comparison of a filename (clearly NOT a path) with something claiming to be a path smells. If DefaultStringFilePath is really a path, it will always fail. If DefaultStringFilePath is really just a name (Appname.xlf), it will succeed for every file when UseLanguageCodeFolders is true, won't it?

---

*[src/L10NSharp/LocalizedStringCache.cs, line 168 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVHQf2G7_0IpRGfol2:-KpVHQf2G7_0IpRGfol3:bt0pwj0) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L168)):*
> ```Smalltalk
> 							Console.WriteLine("WARNING: string ID \"{0}\" already found in \"{1}\".", tu.Id, file);
> 						else
> 							lookupTarget.Add(tu.Id, tu);
> ```

I don't know all the uses of LookupDictsForTransUnits, but could you save space (and later time) by having it just store the localized string for this ID?

---

*[src/L10NSharp/LocalizedStringCache.cs, line 182 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVHlm88nN0hGD2YeTu:-KpVHlm88nN0hGD2YeTv:b-9n47o7) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L182)):*
> ```Smalltalk
> 	#else
> 					// If an error happened reading some localization file other than one we care
> 					// about right now, just ignore it.
> ```

Why do we read ones we don't care about at all? At least until something asks for it (e.g., Bloom sometimes asks for a national language localization).

---

*[src/L10NSharp/LocalizedStringCache.cs, line 292 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVLmsYVSW7TB1_oa7G:-KpVLmsYVSW7TB1_oa7H:b-liyjaq) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L292)):*
> ```Smalltalk
> 				xliffDoc.File.TargetLang = langId;
> 			xliffDoc.File.ProductVersion = OwningManager.AppVersion;
> 			xliffDoc.File.HardLineBreakReplacement = s_literalNewline;
> ```

I've seen something elsewhere about an AmpersandReplacement

---

*[src/L10NSharp/LocalizedStringCache.cs, line 495 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVMZRtVbM03zaPRdyb:-KpVMZRtVbM03zaPRdyc:bzak4qf) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L495)):*
> ```Smalltalk
> 		{
> 			TransUnit tu = DefaultXliffDocument.GetTransUnitForId(id);
> 			return (tu == null || tu.Notes.Count == 0 ? null : tu.Notes[0].Text);
> ```

Why only the first note?

---

*[src/L10NSharp/TransUnitUpdater.cs, line 119 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpV_7Y9db7U-xo5eVYX:-KpV_7Y9db7U-xo5eVYY:bme453n) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L119)):*
> ```Smalltalk
> 					lookupTarget[shortcutId] = tuTargetShortcutKeys;
> 				else if (lookupTarget.ContainsKey(shortcutId))
> 					lookupTarget.Remove(shortcutId);
> ```

Looks like the last seven or so lines are repeated three times. Worth a method?

---

*[src/L10NSharp/TransUnitUpdater.cs, line 163 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpV_RK2IZsNcjxKRZ-q:-KpV_RK2IZsNcjxKRZ-r:b-h0we7z) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L163)):*
> ```Smalltalk
> 			if (tu == null)
> 				return;
> 			if (lookup.ContainsKey(tu.Id))
> ```

The test is redundant; Remove does not care whether the key is present. (Actually it returns a boolean to tell you whether it removed anything.)
Should this set _updated if it removed something?

---

*[src/L10NSharp/TransUnitUpdater.cs, line 181 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVa1pIWEJOZ-PY0wFB:-KpVa1pIWEJOZ-PY0wFC:bobjj5c) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L181)):*
> ```Smalltalk
> 				return;		// empty comment and already no comment in TransUnit
> 			if ((tu.Notes.Count > 0) && (tu.Notes[0].Text == locInfo.Comment))
> 				return;		// exactly the same comment already exists in TransUnit
> ```

Doesn't matter if tu has other Notes? Or can't that happen?

---

*[src/L10NSharp/TransUnitUpdater.cs, line 199 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVaGIN_EfV-WnCZtB2:-KpVaGIN_EfV-WnCZtB3:b-uf0hqb) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L199)):*
> ```Smalltalk
> 			// with users editing interactively and adding tooltips or shortcuts.
> 			Debug.Assert(tuSource == null || tuId == tuSource.Id);
> 			Debug.Assert(tuId.StartsWith(locInfo.Id));
> ```

So if the user can reasonably do something that causes tuSource to be null, that should be handled by non-debug code.

---

*[src/L10NSharp/TransUnitUpdater.cs, line 216 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVahHmfyflrgxr-YYK:-KpVahHmfyflrgxr-YYL:bithh73) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L216)):*
> ```Smalltalk
> 						_updated = true;
> 						tuTarget.RemoveVariant(tuvTarg);
> 						if ((tuTarget.Source == null || tuTarget.Source.Value == null) &&
> ```

Should any of these be nullOrEmpty?

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 98 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVbOaKR-zl5LKGI3D_:-KpVbOaKR-zl5LKGI3Da:b-149flp) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/XLiffUtils/TransUnit.cs#L98)):*
> ```Smalltalk
> 		/// <summary>
> 		/// Gets or sets the priority value.  This is an enumeration which defaults to "DontCare",
> ```

priority comment copied?

---

*[src/L10NSharpTests/LocalizationManagerTests.cs, line 552 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVdr6_hDJnACkYAh6N:-KpVdr6_hDJnACkYAh6O:b-64t0pq) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharpTests/LocalizationManagerTests.cs#L552)):*
> ```Smalltalk
> 				LocalizationManager.LoadedManagers.Clear();
> 				LocalizationManager.SetUILanguage(LocalizationManager.kDefaultLang, false);
> 			}
> ```

Seem to be seeing a lot of similar finally clauses. Would it be worth having a teardown (or setup) function to put the LM in a standard state?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-:-KpVebvWXaMDy3ZTZzha:b-ijlqeb)*

StephenMcConnel:
Review status: 4 of 12 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/L10NSharp/LocalizationManager.cs, line 35 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVBk-l56E1OKI3J3V9:-KpVjKl_rVFz6Y9W4Gm5:b-xutcn3) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizationManager.cs#L35)):*

So English is specifically an exception? Not localization/en/AppName.xlf?
</blockquote></details>

Hmm, I thought it had to be different, but looking at crowdin documentation, their examples show English files in an en subfolder.  So I'll change it to be consistent.  This will simplify one method anyway.  :-)

---

*[src/L10NSharp/LocalizationManager.cs, line 394 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpV9yUkHFuTRBCJfE26:-KpVlhMWMMd2CIeNWZpU:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizationManager.cs#L394)):*

return StringCache.XliffDocuments.Keys.ToList()? (Would also work without the ToList unless the evaluating is expensive or the caller is likely to modify the list.)
</blockquote></details>

Done.

---

*[src/L10NSharp/LocalizationManager.cs, line 407 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVAVL2SGSWTSO6wNhT:-KpVmE0Mo5w-sW313Bmn:b-ypz4q2) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizationManager.cs#L407)):*

Should this have a different implementation if UseLanguageCodeFolders is false? (I see below that a different routine is called in that case...maybe this one could just Assert(UseLanguageCodeFolders)?
Should it have a special case for English (which apparently lives in the root folder)?

</blockquote></details>

Changed to handle English like any other language.  Also added an Assert as suggested.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 109 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVDipoIixMtA2Bb1cX:-KpVqJie65U7f6HlZBn1:b-yrwf43) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L109)):*

It's hard to get this straight without being able to easily navigate to declarations etc. But if defaultXliffDoc and DefaultXliffDocument are different objects, as appears to be the case, I think one of them should have a different name to help the reader figure out what is going on here.
</blockquote></details>

Hmm, I don't think they are different.  The code was adapted from the TMX code, and renaming variables and changing the "document" class to hold only one target language at most changes things enough to make the local variable superfluous as far as I can tell.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 121 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVFL8ciHNhYTREvn0k:-KpVrQ6qjukV7CWJGptR:b-l9z8q1) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L121)):*

This comparison of a filename (clearly NOT a path) with something claiming to be a path smells. If DefaultStringFilePath is really a path, it will always fail. If DefaultStringFilePath is really just a name (Appname.xlf), it will succeed for every file when UseLanguageCodeFolders is true, won't it?
</blockquote></details>

Good catch.  I think "f != OwningManager.DefaultStringFilePath" would be better.  Looking at how the input is generated, I don't think sure any test is needed.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 168 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVHQf2G7_0IpRGfol2:-KpVs7J3gaSQZpTdMJ4A:b-k6pum8) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L168)):*

I don't know all the uses of LookupDictsForTransUnits, but could you save space (and later time) by having it just store the localized string for this ID?
</blockquote></details>

The TransUnit objects are shared between the XliffDocument.File.Body object and the dictionary, so no space would be save by storing the string.  But the same would be true storing the string proper.  It would save a bit of time perhaps.
I've changed the code to store the fast lookup dictionary in the XliffBody class (next to the list of TransUnit objects), and to store the actual string instead of the TransUnit object.  This should be a clearer design for future programmers.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 182 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVHlm88nN0hGD2YeTu:-KpVsXcyf90NMn_BBK6U:b-h7ntit) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L182)):*

Why do we read ones we don't care about at all? At least until something asks for it (e.g., Bloom sometimes asks for a national language localization).
</blockquote></details>

Reason Number 1:  We've always done it that way.  :-)
Reason Number 2:  The code is simpler loading everything at once than having to check whether it needs to be loaded later.
Reason Number 3: The time penalty to load is paid once at startup, so the user won't see a slowdown if something they do accesses localization data that hadn't been loaded.
Deferred loading may be a good idea, but I'd prefer to defer it.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 292 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVLmsYVSW7TB1_oa7G:-KpWQcsbbZSG710NWocJ:b-eyze6k) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L292)):*

I've seen something elsewhere about an AmpersandReplacement
</blockquote></details>

Done.  Added setting that as well.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 495 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVMZRtVbM03zaPRdyb:-KpWN22OvzR_mRwI4yPQ:b-dlnvuo) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L495)):*

Why only the first note?
</blockquote></details>

The specification of the xliff allows multiple notes, but we use only one note as a comment on how to translate the source string.  I'll add a comment to that effect.

---

*[src/L10NSharp/TransUnitUpdater.cs, line 119 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpV_7Y9db7U-xo5eVYX:-KpWRrvQTuS_oGZwYVsO:b-1xx4pa) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L119)):*

Looks like the last seven or so lines are repeated three times. Worth a method?
</blockquote></details>

With other changes, this has been simplified to only 3 lines being repeated three times.  I'll create a method to combine the three lines.  The logic of the first method makes it rather difficult to merely merge the two calls to the second method into its body.

---

*[src/L10NSharp/TransUnitUpdater.cs, line 163 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpV_RK2IZsNcjxKRZ-q:-KpWTlcfVtydfMCVPr0g:b22aezl) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L163)):*

The test is redundant; Remove does not care whether the key is present. (Actually it returns a boolean to tell you whether it removed anything.)
Should this set _updated if it removed something?
</blockquote></details>

This method has been deleted.

---

*[src/L10NSharp/TransUnitUpdater.cs, line 181 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVa1pIWEJOZ-PY0wFB:-KpWU3ezqWKQO7jvfHhZ:bgdskw2) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L181)):*

Doesn't matter if tu has other Notes? Or can't that happen?
</blockquote></details>

That can't happen in our use of comments/notes.  We only allow for the one comment.  Since crowdin likely allows more flexibility, we may want to revisit this decision.

---

*[src/L10NSharp/TransUnitUpdater.cs, line 199 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVaGIN_EfV-WnCZtB2:-KpWUf_cp7BJB-cUwdVb:b52ks6z) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L199)):*

So if the user can reasonably do something that causes tuSource to be null, that should be handled by non-debug code.
</blockquote></details>

I think it's being handled reasonably.  I thought a couple of Asserts as a sanity check during development wouldn't hurt.

---

*[src/L10NSharp/TransUnitUpdater.cs, line 216 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVahHmfyflrgxr-YYK:-KpWUcGyMbNy8fnWha8z:b-yxpibz) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/TransUnitUpdater.cs#L216)):*

Should any of these be nullOrEmpty?
</blockquote></details>

Done.  changed the string comparisons.

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 98 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVbOaKR-zl5LKGI3D_:-KpWVCNRl9ZXQwtKdNrN:b-v1wmrz) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/XLiffUtils/TransUnit.cs#L98)):*

priority comment copied?
</blockquote></details>

yes, because it's the same reasoning for being a string type instead of an enumeration type.

---

*[src/L10NSharpTests/LocalizationManagerTests.cs, line 552 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVdr6_hDJnACkYAh6N:-KpWWlR8BaB_jXLr9gUU:bp7apiv) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharpTests/LocalizationManagerTests.cs#L552)):*

Seem to be seeing a lot of similar finally clauses. Would it be worth having a teardown (or setup) function to put the LM in a standard state?
</blockquote></details>

Done.  added a TearDown method.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17)*

StephenMcConnel:
Review status: 4 of 12 files reviewed at latest revision, 17 unresolved discussions.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 107 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVECBDV8Nw4UjymHk2:-KpWfudmpIxBOP__0fHu:bn5tc0f) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L107)):*

I wonder whether we should encapsulate the File.Body bit by giving XLiffDocument an AddTransUnitOrVariantFromExisting method? One day we might want to make the in-memory object store translations in some form more compact than an XML document and be sorry we let this implementation detail escape. (I see below that File.Body. Whatever is a common pattern, and was even before this PR. So probably not something to fix now. A thought to consider as you think about further enhancement.)
</blockquote></details>

I'd rather postpone this design issue for now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17)*

JohnThomson:
Reviewed 8 of 8 files at r2.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---

*[src/L10NSharp/LocalizationManager.cs, line 35 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVBk-l56E1OKI3J3V9:-KpX52CBUohNBrqQIcNU:b-lwve5) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizationManager.cs#L35)):*

Hmm, I thought it had to be different, but looking at crowdin documentation, their examples show English files in an en subfolder.  So I'll change it to be consistent.  This will simplify one method anyway.  :-)
</blockquote></details>

I like this change, but I have a vague recollection of a method that needed to enumerate all the languages EXCEPT the default, and that it might have worked because English didn't fit the pattern. So you might take a quick look for that.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 107 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVECBDV8Nw4UjymHk2:-KpX4L0jC9KY5GN7tFwU:blyphgp) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L107)):*

I'd rather postpone this design issue for now.
</blockquote></details>

No problem.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 182 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVHlm88nN0hGD2YeTu:-KpX5gsUv3Ai4u0cXu9d:bgf9vrp) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizedStringCache.cs#L182)):*

Reason Number 1:  We've always done it that way.  :-)
Reason Number 2:  The code is simpler loading everything at once than having to check whether it needs to be loaded later.
Reason Number 3: The time penalty to load is paid once at startup, so the user won't see a slowdown if something they do accesses localization data that hadn't been loaded.
Deferred loading may be a good idea, but I'd prefer to defer it.
</blockquote></details>

That's fine. Sorry to bug you about it.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 92 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpX4CEKsOYJoO9j7amY:-KpX4CEKsOYJoO9j7amZ:bd820df) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a5e269c30cdd89dbe6bea3cfa0e0a2891e56b754/src/L10NSharp/LocalizedStringCache.cs#L92)):*
> ```Smalltalk
> 			// Fill in the fast lookup dictionary.
> 			foreach (var tu in DefaultXliffDocument.File.Body.TransUnits)
> 				DefaultXliffDocument.File.Body.TranslationsById.Add(tu.Id, tu.GetVariantForLang(LocalizationManager.kDefaultLang).Value);
> ```

Would it make sense for the document object to know how to do this itself now? Or even do it automatically as part of Read()?

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 98 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVbOaKR-zl5LKGI3D_:-KpX6ZlSO18CvrbxZJPh:b-853wb9) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/XLiffUtils/TransUnit.cs#L98)):*

yes, because it's the same reasoning for being a string type instead of an enumeration type.
</blockquote></details>

I was too cryptic. No problem copying the appropriate part of a similar comment, but shouldn't this one begin "Gets or sets the category"? (I'm curious what the other standard values are...if any...)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17)*

StephenMcConnel:
Review status: 9 of 12 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/L10NSharp/LocalizationManager.cs, line 35 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVBk-l56E1OKI3J3V9:-KpX8EPsfYyKYrkWObfK:bhu4cn4) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/LocalizationManager.cs#L35)):*

I like this change, but I have a vague recollection of a method that needed to enumerate all the languages EXCEPT the default, and that it might have worked because English didn't fit the pattern. So you might take a quick look for that.
</blockquote></details>

I think you're referring to the method providing files to merge in after loading the default (English) file (LocalizationManager.XliffFilenamesToAddToCache).  It already checked for the default language, and did (does) not return that file.  :-)

---

*[src/L10NSharp/LocalizedStringCache.cs, line 92 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpX4CEKsOYJoO9j7amY:-KpXE4ccglOrggiLtWfk:bnt5pb6) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a5e269c30cdd89dbe6bea3cfa0e0a2891e56b754/src/L10NSharp/LocalizedStringCache.cs#L92)):*

Would it make sense for the document object to know how to do this itself now? Or even do it automatically as part of Read()?
</blockquote></details>

It's been changed to be part of Read(), but the caches need adjusting for "orphans" that have merely changed ids in the translation files.

---

*[src/L10NSharp/XLiffUtils/TransUnit.cs, line 98 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-KpVbOaKR-zl5LKGI3D_:-KpXFUjzeFufITcv35Vi:bj2a0vg) ([raw file](https://github.com/sillsdev/l10nsharp/blob/2c9b22bce129ca7aeb1979936ac2ff73a4cedb88/src/L10NSharp/XLiffUtils/TransUnit.cs#L98)):*

I was too cryptic. No problem copying the appropriate part of a similar comment, but shouldn't this one begin "Gets or sets the category"? (I'm curious what the other standard values are...if any...)
</blockquote></details>

Fixed it.  Also added pointers to the LocalizationPriority and LocalizationCategory enums in the comments.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17)*

JohnThomson:
:lgtm:

---

Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/17#-:-Kp_OQhYrzk_GKNvf7sn:bnfp4nl)*

PR #22, opened 2017-08-18
Fix hanging bug on Linux scanning libpalaso assemblies

StephenMcConnel:
This fixes http://issues.bloomlibrary.org/youtrack/issue/BL-4840

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/22)

### DIFF
--- a/src/L10NSharp/CodeReader/StringExtractor.cs
+++ b/src/L10NSharp/CodeReader/StringExtractor.cs
@@ -208,12 +208,19 @@ namespace L10NSharp.CodeReader
 			methodsInType.AddRange(type.GetConstructors(BindingFlags.Static |
 				BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic));
 
-			// The DeclaredOnly flag ensures we do not include public methods from superclasses of this subtype.
-			// If the superclass is in a requested namespace, we will collect it from the superclass directly.
-			// But this ensures we don't collect it if the superclass is in a namespace which wasn't requested.
-			methodsInType.AddRange(type.GetMethods(BindingFlags.Static |
-				BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly));
-
+			try
+			{
+				// The DeclaredOnly flag ensures we do not include public methods from superclasses of this subtype.
+				// If the superclass is in a requested namespace, we will collect it from the superclass directly.
+				// But this ensures we don't collect it if the superclass is in a namespace which wasn't requested.
+				methodsInType.AddRange(type.GetMethods(BindingFlags.Static |
+					BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly));
+			}
+			catch (TypeLoadException)
+			{
+				// Caused by assemblies that have odd runtime loading problems (e.g. SIL.Windows.Forms.Keyboarding). Ignore.
+				return;
+			}
 			foreach (var method in methodsInType)
 			{
 				try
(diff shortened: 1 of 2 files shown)


src/L10NSharp/CodeReader/StringExtractor.cs
```diff
@@ -208,12 +208,19 @@ private void FindLocalizedStringsInType(Type type)
...
+			{
+				// The DeclaredOnly flag ensures we do not include public methods from superclasses of this subtype.
+				// If the superclass is in a requested namespace, we will collect it from the superclass directly.
+				// But this ensures we don't collect it if the superclass is in a namespace which wasn't requested.
+				methodsInType.AddRange(type.GetMethods(BindingFlags.Static |
+					BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic | BindingFlags.DeclaredOnly));
+			}
+			catch (TypeLoadException)
+			{
+				// Caused by assemblies that have odd runtime loading problems (e.g. SIL.Windows.Forms.Keyboarding). Ignore.
```
hatton: My concern is that a developer might be counting on that type being scanned, but now it will silently not be. I can't immediately think of a solution though. Maybe you will.

StephenMcConnel:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/L10NSharp/CodeReader/StringExtractor.cs, line 221 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/22#-KrqLaXWMaFGvj-iWX98-r1-221:-KrqOQuKz68aHfonPlRp:b-eh1vzg) ([raw file](https://github.com/sillsdev/l10nsharp/blob/4b19bcd708a65e8b6b6e5f73bd74bd8f2934f6d5/src/L10NSharp/CodeReader/StringExtractor.cs#L221)):*
<details><summary><i>Previously, hatton (John Hatton) wrote…</i></summary><blockquote>

My concern is that a developer might be counting on that type being scanned, but now it will silently not be. I can't immediately think of a solution though. Maybe you will.
</blockquote></details>

I thought this is code that we've discussed omitting entirely in the future?
The other try-catches in this method worked silently.  I suppose I could add Debug.WriteLine or Console.WriteLine messages in all those places, but I don't know that those would ever be noticed anyway.  These errors have almost always been Linux specific, so depending on Windows output would probably still be fairly robust.  The Mono runtime is just pickier in some ways and quicker to throw errors than the .Net runtime.  Of course the Windows output won't have any Linux specific message strings, so that's not a perfect solution.  (and vice-versa)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/22)*

StephenMcConnel:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/L10NSharp/CodeReader/StringExtractor.cs, line 221 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/22#-KrqLaXWMaFGvj-iWX98-r1-221:-KrqQ1OuuXeReFZGDj54:b2o5ikm) ([raw file](https://github.com/sillsdev/l10nsharp/blob/4b19bcd708a65e8b6b6e5f73bd74bd8f2934f6d5/src/L10NSharp/CodeReader/StringExtractor.cs#L221)):*

I thought this is code that we've discussed omitting entirely in the future?
The other try-catches in this method worked silently.  I suppose I could add Debug.WriteLine or Console.WriteLine messages in all those places, but I don't know that those would ever be noticed anyway.  These errors have almost always been Linux specific, so depending on Windows output would probably still be fairly robust.  The Mono runtime is just pickier in some ways and quicker to throw errors than the .Net runtime.  Of course the Windows output won't have any Linux specific message strings, so that's not a perfect solution.  (and vice-versa)
</blockquote></details>

I created a new issue to deal with logging these exceptions somehow. (BL-4902)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/22)*

PR #24, opened 2017-11-08
Allow 3-letter ISO codes for localization languages

StephenMcConnel:
This is on the xliff branch.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/24)

### DIFF
--- a/src/L10NSharp/LocalizationManager.cs
+++ b/src/L10NSharp/LocalizationManager.cs
@@ -394,7 +394,7 @@ namespace L10NSharp
 
 		/// <summary>
 		/// If the given file exists, return its parent folder name as a language tag if it
-		/// appears to be valid (2 characters long or "zh-CN").  Otherwise return null.
+		/// appears to be valid (2 or 3 letters long or "zh-CN").  Otherwise return null.
 		/// </summary>
 		private static string GetLanguageTagFromFilePath(string xliffFile)
 		{
@@ -402,7 +402,7 @@ namespace L10NSharp
 			if (File.Exists(xliffFile))
 			{
 				var langId = Path.GetFileName(Path.GetDirectoryName(xliffFile));
-				if ((langId.Length == 2 || langId == "zh-CN"))
+				if (System.Text.RegularExpressions.Regex.IsMatch(langId, "[a-z]{2,3}") || langId == "zh-CN")
 					return langId;
 			}
 			return null;


JohnThomson:
:lgtm: with one question

---

Reviewed 1 of 1 files at r1.

---

*[src/L10NSharp/LocalizationManager.cs, line 405 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/24#-KyWGXFKECGrEFwynOQv:-KyWGXFKECGrEFwynOQw:bn20z4l) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e7348f7aace3191be5d3ca1eee8a4d7610840d3d/src/L10NSharp/LocalizationManager.cs#L405)):*
> ```Smalltalk
> 			{
> 				var langId = Path.GetFileName(Path.GetDirectoryName(xliffFile));
> 				if (System.Text.RegularExpressions.Regex.IsMatch(langId, "[a-z]{2,3}") || langId == "zh-CN")
> ```

This relaxes things as to length, but tightens them in requiring lower-case alphabetic ascii characters. I think that's what ISO requires, but just wanted to check that you've thought carefully about any problems it might cause.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/24#-:-KyWGoac6HWYPzDQGxNS:b-cn9p1x)*

StephenMcConnel:

---

*[src/L10NSharp/LocalizationManager.cs, line 405 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/24#-KyWGXFKECGrEFwynOQv:-KyWHLTUC2DINx1hQl0b:b-iroznw) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e7348f7aace3191be5d3ca1eee8a4d7610840d3d/src/L10NSharp/LocalizationManager.cs#L405)):*

This relaxes things as to length, but tightens them in requiring lower-case alphabetic ascii characters. I think that's what ISO requires, but just wanted to check that you've thought carefully about any problems it might cause.
</blockquote></details>

Yes, ISO has 2-letter and 3-letter codes for languages, and the codes all consists of lowercase English/Latin letters without diacritics.  We use the ISO code for the directory name.  Before now, only 2-letter codes have been needed (these are available only for "major" languages), but there are a couple of languages with 3-letter codes in the pipeline for localization.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/24)*

PR #26, opened 2017-11-14
Create ExtractXliff command line program

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/26)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ output/
 obj/
 *.suo
 *.user
+L10NSharp.userprefs
 bin/
 *.ncrunchsolution
 *Copy*.bat
--- a/L10NSharp.sln
+++ b/L10NSharp.sln
@@ -1,11 +1,14 @@
-
-Microsoft Visual Studio Solution File, Format Version 11.00
-# Visual Studio 2010
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
+MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "L10NSharp", "src\L10NSharp\L10NSharp.csproj", "{8EE0895A-7B9D-4B49-8F73-9F1270F1E583}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "L10NSharpTests", "src\L10NSharpTests\L10NSharpTests.csproj", "{BCE5B569-057C-4D7E-832E-E44A2DA705AC}"
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SampleApp", "src\SampleApp\SampleApp.csproj", "{58923B30-FD84-4BCC-85E0-607DCA7E7C95}"
+EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ExtractXliff", "src\ExtractXliff\ExtractXliff.csproj", "{E0666C78-B8DC-4232-952C-753940D54921}"
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
@@ -47,6 +50,18 @@ Global
 		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|Mixed Platforms.Build.0 = Release|x86
 		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|x86.ActiveCfg = Release|x86
 		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|x86.Build.0 = Release|x86
+		{E0666C78-B8DC-4232-952C-753940D54921}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Debug|Mixed Platforms.Build.0 = Debug|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Debug|x86.Build.0 = Debug|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Release|Any CPU.Build.0 = Release|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Release|Mixed Platforms.Build.0 = Release|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Release|x86.ActiveCfg = Release|Any CPU
+		{E0666C78-B8DC-4232-952C-753940D54921}.Release|x86.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/src/ExtractXliff/App.config
+++ b/src/ExtractXliff/App.config
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.6.1"/>
+    </startup>
+</configuration>
--- a/src/ExtractXliff/ExtractXliff.csproj
+++ b/src/ExtractXliff/ExtractXliff.csproj
@@ -1,0 +1,70 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProjectGuid>{E0666C78-B8DC-4232-952C-753940D54921}</ProjectGuid>
+    <OutputType>Exe</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>ExtractXliff</RootNamespace>
+    <AssemblyName>ExtractXliff</AssemblyName>
+    <TargetFrameworkVersion>v4.6.1</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
+    <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+    <TargetFrameworkProfile />
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>..\..\output\Debug\</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <Externalconsole>true</Externalconsole>
+    <Commandlineparameters>-v -n Bloom -o Bloom.exe -p 3.1.000.0 -x /tmp/Bloom.xlf -b /d/github/BloomV4.0/DistFiles/localization/en/Bloom.xlf /d/github/BloomV4.0/output/Debug/Bloom.exe</Commandlineparameters>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>..\..\output\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Drawing" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="Program.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="App.config" />
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
+       Other similar extension points exist, see Microsoft.Common.targets.
+  <Target Name="BeforeBuild">
+  </Target>
+  <Target Name="AfterBuild">
+  </Target>
+  -->
+  <ItemGroup>
+    <ProjectReference Include="..\L10NSharp\L10NSharp.csproj">
+      <Project>{8EE0895A-7B9D-4B49-8F73-9F1270F1E583}</Project>
+      <Name>L10NSharp</Name>
+    </ProjectReference>
+  </ItemGroup>
+</Project>
--- a/src/ExtractXliff/Program.cs
+++ b/src/ExtractXliff/Program.cs
@@ -1,0 +1,456 @@
+﻿// Copyright (c) 2017 SIL International
+// This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
+using System;
+using System.Collections.Generic;
+using System.Drawing;
+using System.Text;
+using System.Threading.Tasks;
+using System.Reflection;
+using System.Linq;
+using System.Diagnostics;
+using System.Resources;
+using System.Globalization;
+using System.Collections;
+using L10NSharp;
+using L10NSharp.CodeReader;
+using L10NSharp.XLiffUtils;
+
+namespace ExtractXliff
+{
+	/// <summary>
+	/// This program uses the L10NSharp code for extracting static strings from one or
+	/// more C# assemblies (either .dll or .exe).  It requires command line arguments
+	/// to set the internal XLIFF file element "original" attribute, to set the
+	/// namespace beginning(s), and to set the output XLIFF filename.  It also requires
+	/// one or more assembly files to be specified on the command line.  There are
+	/// also optional command line arguments for specifying the XLIFF file element
+	/// "datatype" attribute, the XLIFF file element "product-version" attribute, and
+	/// an existing XLIFF file to merge from after reading everything from the input
+	/// assemblies.
+	/// </summary>
+	class Program
+	{
+		delegate bool ProcessOption(string[] args, ref int i);
+
+		static List<string> _namespaces = new List<string>();	// namespace beginning(s) (-n)
+		static string _xliffOutputFilename;			// new XLIFF output file (-x)
+		static string _fileDatatype;				// file element attribute value (-d)
+		static string _fileOriginal;				// file element attribute value (-o)
+		static string _fileProductVersion;			// file element attribute value (-p)
+		static string _baseXliffFilename;			// input file that provides existing data to merge (-b)
+		static bool _verbose;						// verbose console output (-v)
+		static bool _doneWithOptions;				// flag that either "--" or first assembly filename seen already
+		static List<string> _assemblyFiles = new List<string>();	// input assembly file(s)
+
+		const string kDefaultLangId = "en";
+		const string kDefaultNewlineReplacement = "\\n";
+		const string kDefaultAmpersandReplacement = "|amp|";
+
+		static void Main(string[] args)
+		{
+			// Parse and check the command line arguments.
+			if (!ParseOptions(args))
+			{
+				Usage();
+				return;
+			}
+
+			// Load the input assemblies so that they can be scanned.
+			List<Assembly> assemblies = new List<Assembly>();
+			foreach (var file in _assemblyFiles)
+			{
+				var asm = Assembly.LoadFile(file);
+				if (asm != null)
+					assemblies.Add(asm);
+			}
+
+			// Scan the input assemblies for localizable strings.
+			var extractor = new StringExtractor();
+			extractor.ExternalAssembliesToScan = assemblies.ToArray();
+			var localizedStrings = extractor.DoExtractingWork(_namespaces.ToArray(), null);
+
+			// The arguments to this constructor don't really matter much as they're used internally by
+			// L10NSharp for reasons that may not percolate out to xliff.  We just need a LocalizationManager
+			// to feed into the constructor the LocalizedStringCache that does some heavy lifting for us in
+			// creating the XliffDocument from the newly extracted localized strings.
+			var lm = new LocalizationManager(_fileOriginal, _fileOriginal, _fileProductVersion);
+			var stringCache = new LocalizedStringCache(lm, false);
+			foreach (var locInfo in localizedStrings)
+				stringCache.UpdateLocalizedInfo(locInfo);
+
+			// Get the newly loaded static strings (in newDoc) and the baseline XLIFF (in baseDoc).
+			var newDoc = stringCache.XliffDocuments[kDefaultLangId];
+			var baseDoc = LoadBaselineAndCompare(newDoc);
+
+			// Save the results to the output file, merging in data from the baseline XLIFF if one was specified.
+			MergeAndSaveXliffDataToFile(newDoc, baseDoc);
+		}
+
+		/// <summary>
+		/// Saves the merged XLIFF data to the output file.
+		/// </summary>
+		private static void MergeAndSaveXliffDataToFile(XLiffDocument xliffNew, XLiffDocument xliffOld)
+		{
+			// xliffNew has the data found in the current scan.
+			// xliffOld is that data from the (optional) input baseline XLIFF file.
+			// xliffOutput is the data that is actually written to the designated output XLIFF file.  It combines
+			//   data from both xliffNew and xliffOld.
+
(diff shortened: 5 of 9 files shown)


JohnThomson:
Feels like I'm being way too critical, especially when this is basically a simple utility program that probably won't get much maintenance. Or maybe I just don't understand xliff well enough to be reviewing this. Feel free to push back on changes that don't seem worth the effort.

---

Review status: 0 of 10 files reviewed at latest revision, 14 unresolved discussions.

---

*[src/ExtractXliff/Program.cs, line 21 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywVbxeRjc0Ac1E6lEv:-KywVbxeRjc0Ac1E6lEw:b-hcem3n) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L21)):*
> ```Smalltalk
> 	{
> 		static List<string> _namespaces = new List<string>();
> 		static string _xliffFilename;
> ```

This is in fact I believe the OUTPUT xliff filename. Unfortunately it's inherent in the nature of what this program does that we have both a 'new' and an 'output' xliff file; the two are easily confused, I find, since 'new' strongly suggests an output. I think it would be helpful to have a class comment which clearly identifies the three xliff files involved in the process and a consistent naming convention for them which is followed throughout (including using a name here that makes it clear which of the three it is).

---

*[src/ExtractXliff/Program.cs, line 23 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywUnZCr3p3ANQP9IYz:-KywUnZCr3p3ANQP9IZ-:b-2im62g) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L23)):*
> ```Smalltalk
> 		static string _xliffFilename;
> 		static string _datatype;
> 		static string _original;
> ```

If I'm understanding right, _original is (a) the value of the -o argument; (b) the name of the 'new' .xlf file (c) the id and name of the LocalizationManager, and (d) an attribute value of the file element in the new .xlf file. It's not at all obvious why all of those should be the same, or why the variable for the third of the xliff files involved in the process isn't <something>xliffFilename. I think at least a comment is indicated, especially since 'original' strongly suggests to me the OLD file (what you're calling base).

---

*[src/ExtractXliff/Program.cs, line 34 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywTWf9Jdps0IwKh_LJ:-KywTWf9Jdps0IwKh_LK:bzffuqs) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L34)):*
> ```Smalltalk
> 		const string kDefaultAmpersandReplacement = "|amp|";
> 
> 		static void Main(string[] args)
> ```

This method is basically three steps: parse the options, check the options, do the work. Two of the steps are encapsulated in methods while one is spelled out here, making it seem as if the main point of the program is to check the options. It's not vital, but it would feel neater to make CheckOptions a method too.

---

*[src/ExtractXliff/Program.cs, line 53 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywQtoq1kzLyRfl1J7i:-KywQtoq1kzLyRfl1J7j:b3m0sjx) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L53)):*
> ```Smalltalk
> 				extractor.ExternalAssembliesToScan = assemblies.ToArray();
> 				IEnumerable<LocalizingInfo> localizedStrings = extractor.DoExtractingWork(_namespaces.ToArray(), null);
> 				var lm = new LocalizationManager(_original, _original, _productVersion);
> ```

This will presumably crash if they didn't provide the _orginal argument, but it would be more consistent to give a nice message with Usage, wouldn't it? Also it's not clear from Usage that original is used for more than just an attribute of an element in the output file

---

*[src/ExtractXliff/Program.cs, line 62 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywNqMAOnjzKWcI8ndt:-KywNqMAOnjzKWcI8ndu:b-ualayz) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L62)):*
> ```Smalltalk
> 				{
> 					baseDoc = XLiffDocument.Read(_baseXliffFilename);
> 					if (baseDoc.File.SourceLang != newDoc.File.SourceLang && baseDoc.File.SourceLang != kDefaultLangId)
> ```

Is it really OK that, say, the base document maps English to French and the new one maps French to Spanish? (Not likely, of course...)

---

*[src/ExtractXliff/Program.cs, line 68 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywNdRbycnNNHGQhhNV:-KywNdRbycnNNHGQhhNW:b6vl1aq) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L68)):*
> ```Smalltalk
> 						return;
> 					}
> 					if (baseDoc.File.Original != newDoc.File.Original && baseDoc.File.Original != _original)
> ```

I'm not clear what original is, but it is it really OK that they are both the same but different from what the output file will use?

---

*[src/ExtractXliff/Program.cs, line 73 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywNFfGt_SdKBGsiVUN:-KywNFfGt_SdKBGsiVUO:b-a4xqlv) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L73)):*
> ```Smalltalk
> 							baseDoc.File.Original, newDoc.File.Original);
> 					}
> 					if (baseDoc.File.DataType != newDoc.File.DataType && baseDoc.File.DataType != _datatype)
> ```

Is it really OK that they both have the same DataType which is different from the datatype that will be applied to the merged file?

---

*[src/ExtractXliff/Program.cs, line 78 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywTBfYUqBlPE0Qi5Ov:-KywTBfYUqBlPE0Qi5Ow:bt8hapy) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L78)):*
> ```Smalltalk
> 							baseDoc.File.DataType, newDoc.File.DataType);
> 					}
> 					if (baseDoc.File.ProductVersion != newDoc.File.ProductVersion && baseDoc.File.ProductVersion != _productVersion)
> ```

Again, it's OK if the old and new files agree on the product version, even though it's different from the one we will write to the merged file?

---

*[src/ExtractXliff/Program.cs, line 94 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywXFtBCCzsooWWw-pJ:-KywXFtBCCzsooWWw-pK:btuxpbw) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L94)):*
> ```Smalltalk
> 			if (!String.IsNullOrEmpty(_productVersion))
> 				xliffOutput.File.ProductVersion = _productVersion;
> 			xliffOutput.File.HardLineBreakReplacement = kDefaultNewlineReplacement;
> ```

Is it possible/a problem if one or both of the input files use some other replacement strings?

---

*[src/ExtractXliff/Program.cs, line 132 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywJzUHpSsjWOMqx9yL:-KywJzUHpSsjWOMqx9yM:b-40qfxz) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L132)):*
> ```Smalltalk
> 						if (tuOld.Dynamic)
> 						{
> 							++wrongDynamicFlagCount;
> ```

How do we know old dynamic was wrong? Are we assuming the new xliff is the output of a process that won't find dynamic strings?

---

*[src/ExtractXliff/Program.cs, line 255 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywL3rX7P5YhI7hFwqu:-KywL3rX7P5YhI7hFwqv:b-rqspjj) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L255)):*
> ```Smalltalk
> 			if (error)
> 				return false;
> 			if (args[i] == "-n" || "--namespace".StartsWith(args[i]))
> ```

Is it intentional that an argument that is simply -- or even just a single hyphen will be taken as a namespace argument? (Same with other methods, except they won't get the chance).
Is it intentional to be case-sensitive?

---

*[src/ExtractXliff/Program.cs, line 369 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywMEhoMYO68CJvpBQZ:-KywMEhpT4eWSK4GjM6U:bcvjep2) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L369)):*
> ```Smalltalk
> 			}
> 			return false;
> 		}
> ```

Seems nearly all these methods do much the same thing and could be a single method taking a short name, long name, and an Action<string> that does something with the following argument.

---

*[src/L10NSharp/LocalizationManager.cs, line 244 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywSC3bJVtpn02DQ9mJ:-KywSC3bJVtpn02DQ9mK:b6r3al6) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/L10NSharp/LocalizationManager.cs#L244)):*
> ```Smalltalk
> 			Name = appName;
> 			AppVersion = appVersion;
> 		}
> ```

It appears the only caller of this is the new call above which passes _original for the first two arguments. I'm not sure what Id and Name are used for, but it seems some comment might be called for here or where that call happens to explain why it is a good value for both. Or maybe in this special usage the Id and Name just don't matter much? Anyway some clarification would be nice.

---

*[src/L10NSharp/CodeReader/StringExtractor.cs, line 180 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywRrf_1ORzcy78UUJ-:-KywRrf_1ORzcy78UUJ0:b-axt28) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/L10NSharp/CodeReader/StringExtractor.cs#L180)):*
> ```Smalltalk
> 		/// <summary>
> 		/// Gets a list of all assemblies referenced by the entry assembly
> ```

Seems some new commenting is needed here and/or on ExternalAssembliesToScan to make it clear that if that is used we don't scan the entry assembly or other loaded assemblies.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-:-KywXcpLEksz1x-U3j4k:bxfxps7)*

JohnThomson:
Reviewed 10 of 10 files at r1.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26)*

StephenMcConnel:
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---

*[src/ExtractXliff/Program.cs, line 23 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywUnZCr3p3ANQP9IYz:-Kz5f2zV5kE5DXYhxYZd:bj1or3v) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L23)):*

If I'm understanding right, _original is (a) the value of the -o argument; (b) the name of the 'new' .xlf file (c) the id and name of the LocalizationManager, and (d) an attribute value of the file element in the new .xlf file. It's not at all obvious why all of those should be the same, or why the variable for the third of the xliff files involved in the process isn't <something>xliffFilename. I think at least a comment is indicated, especially since 'original' strongly suggests to me the OLD file (what you're calling base).
</blockquote></details>

_original is not the name of the "new" .xlf file.   I've renamed the class variables, and put comments on each of them to try to clarify things.  _original is used as the appId and appName of the localization manager and as the XLIFF file element attribute value.  The use in the localization manager is perhaps pragmatic, but it fits in with the initialization calls in the Bloom code.

---

*[src/ExtractXliff/Program.cs, line 34 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywTWf9Jdps0IwKh_LJ:-Kz5dyBDgsR5X5HHBMNi:b-68f8qp) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L34)):*

This method is basically three steps: parse the options, check the options, do the work. Two of the steps are encapsulated in methods while one is spelled out here, making it seem as if the main point of the program is to check the options. It's not vital, but it would feel neater to make CheckOptions a method too.
</blockquote></details>

The program structure obviously isn't clear enough.  I've add some comments and split off another method from Main.  I've also added a class comment and method comments throughout.

---

*[src/ExtractXliff/Program.cs, line 53 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywQtoq1kzLyRfl1J7i:-Kz5gF91w2i6DdvXLqlQ:b-n2ao87) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L53)):*

This will presumably crash if they didn't provide the _orginal argument, but it would be more consistent to give a nice message with Usage, wouldn't it? Also it's not clear from Usage that original is used for more than just an attribute of an element in the output file
</blockquote></details>

Providing the _original argument is enforced by ParseOptions.  The program will display the Usage() message and exit if it isn't provided.

---

*[src/ExtractXliff/Program.cs, line 62 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywNqMAOnjzKWcI8ndt:-Kz5g_MwDqMUfhrHANrJ:b-72jzyz) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L62)):*

Is it really OK that, say, the base document maps English to French and the new one maps French to Spanish? (Not likely, of course...)
</blockquote></details>

No, the check really is enforcing English as the only valid source-language attribute value.

---

*[src/ExtractXliff/Program.cs, line 68 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywNdRbycnNNHGQhhNV:-Kz5i2Z2GVLgzaLQdC5P:b-y5tld1) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L68)):*

I'm not clear what original is, but it is it really OK that they are both the same but different from what the output file will use?
</blockquote></details>

I've simplified the checks to better mirror what will actually appear in the output file.

---

*[src/ExtractXliff/Program.cs, line 73 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywNFfGt_SdKBGsiVUN:-Kz5i9f4S6yyG_sKrVXT:b-y5tld1) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L73)):*

Is it really OK that they both have the same DataType which is different from the datatype that will be applied to the merged file?
</blockquote></details>

I've simplified the checks to better mirror what will actually appear in the output file.

---

*[src/ExtractXliff/Program.cs, line 78 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywTBfYUqBlPE0Qi5Ov:-Kz5iAx9MW4gOA9Bzkf8:b-y5tld1) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L78)):*

Again, it's OK if the old and new files agree on the product version, even though it's different from the one we will write to the merged file?
</blockquote></details>

I've simplified the checks to better mirror what will actually appear in the output file.

---

*[src/ExtractXliff/Program.cs, line 94 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywXFtBCCzsooWWw-pJ:-Kz5iCsbVVj8YdhF5FyZ:b2ovcq6) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L94)):*

Is it possible/a problem if one or both of the input files use some other replacement strings?
</blockquote></details>

no.  At the moment, there's no way to modify these replacement strings anywhere, and I'm not sure they're even being used in practice.  They may be relics of the old TMX code, but I didn't want to change everything at once.

---

*[src/ExtractXliff/Program.cs, line 132 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywJzUHpSsjWOMqx9yL:-Kz5iTG23ncVvFBrwrVT:b-5xw7ky) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L132)):*

How do we know old dynamic was wrong? Are we assuming the new xliff is the output of a process that won't find dynamic strings?
</blockquote></details>

Yes, the new values (from newDoc) are found by the static scan process implemented by L10NSharp.  So nothing it finds can be "dynamic" (unless of course I have no clue what "dynamic" means in the L10NSharp domain).

---

*[src/ExtractXliff/Program.cs, line 255 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywL3rX7P5YhI7hFwqu:-Kz5it2OF9oVb51QLFjx:bd2h1ry) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L255)):*

Is it intentional that an argument that is simply -- or even just a single hyphen will be taken as a namespace argument? (Same with other methods, except they won't get the chance).
Is it intentional to be case-sensitive?
</blockquote></details>

It is intentional to be case-sensitve.  If the user types garbage on the command line that passes the minimal syntax checking of the code, then the output is likely to be garbage and possibly empty.  I'm not trying to prevent all errors, just ones that will crash the program immediately.

---

*[src/ExtractXliff/Program.cs, line 369 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywMEhoMYO68CJvpBQZ:-Kz5jQXRCMPFLgzkmpxx:b-7rl0jd) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L369)):*

Seems nearly all these methods do much the same thing and could be a single method taking a short name, long name, and an Action<string> that does something with the following argument.
</blockquote></details>

done.

---

*[src/L10NSharp/LocalizationManager.cs, line 244 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywSC3bJVtpn02DQ9mJ:-Kz5r26uXH8uQKvX7xnO:b-7rl0jd) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/L10NSharp/LocalizationManager.cs#L244)):*

It appears the only caller of this is the new call above which passes _original for the first two arguments. I'm not sure what Id and Name are used for, but it seems some comment might be called for here or where that call happens to explain why it is a good value for both. Or maybe in this special usage the Id and Name just don't matter much? Anyway some clarification would be nice.
</blockquote></details>

done.

---

*[src/L10NSharp/CodeReader/StringExtractor.cs, line 180 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywRrf_1ORzcy78UUJ-:-Kz5p7CPv9L3MeUTuv_y:b-7rl0jd) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/L10NSharp/CodeReader/StringExtractor.cs#L180)):*

Seems some new commenting is needed here and/or on ExternalAssembliesToScan to make it clear that if that is used we don't scan the entry assembly or other loaded assemblies.
</blockquote></details>

done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26)*

JohnThomson:
Much better. Two minor points still. Feel free to merge if I'm off track with them.

---

Review status: 6 of 10 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/ExtractXliff/Program.cs, line 23 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywUnZCr3p3ANQP9IYz:-Kz5umJXsBlRvP_ypIPc:b-em55yq) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L23)):*

_original is not the name of the "new" .xlf file.   I've renamed the class variables, and put comments on each of them to try to clarify things.  _original is used as the appId and appName of the localization manager and as the XLIFF file element attribute value.  The use in the localization manager is perhaps pragmatic, but it fits in with the initialization calls in the Bloom code.
</blockquote></details>

My bad. Somewhere I got the idea that the new data came from an xliff file, but of course it's really from the C# assemblies.

---

*[src/ExtractXliff/Program.cs, line 34 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywTWf9Jdps0IwKh_LJ:-Kz5zNDtjTJ0CbRg5-oi:bo3kj57) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L34)):*

The program structure obviously isn't clear enough.  I've add some comments and split off another method from Main.  I've also added a class comment and method comments throughout.
</blockquote></details>

Makes much more sense now (especially once I got straight that there was only one input xliff file).

---

*[src/ExtractXliff/Program.cs, line 53 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywQtoq1kzLyRfl1J7i:-Kz6-7T-NFtVCzr_GOff:buq1qyh) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L53)):*

Providing the _original argument is enforced by ParseOptions.  The program will display the Usage() message and exit if it isn't provided.
</blockquote></details>

I see that now I look closely.

---

*[src/ExtractXliff/Program.cs, line 132 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywJzUHpSsjWOMqx9yL:-Kz5vo1jg7PL03MSYDMx:b-4b0djm) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L132)):*

Yes, the new values (from newDoc) are found by the static scan process implemented by L10NSharp.  So nothing it finds can be "dynamic" (unless of course I have no clue what "dynamic" means in the L10NSharp domain).
</blockquote></details>

Of course, it's not an assumption, you do the process here! I must have been really tired that evening.

---

*[src/ExtractXliff/Program.cs, line 255 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywL3rX7P5YhI7hFwqu:-Kz5wLzRv44-AyRJJ4Tl:b-at0gi0) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L255)):*

It is intentional to be case-sensitve.  If the user types garbage on the command line that passes the minimal syntax checking of the code, then the output is likely to be garbage and possibly empty.  I'm not trying to prevent all errors, just ones that will crash the program immediately.
</blockquote></details>

I still don't think it's OK that "--", which has a specific purpose of terminating the arguments list, will instead be taken as matching any argument not already looked for (since e.g., "--namespace".startsWith("--") is true)

---

*[src/ExtractXliff/Program.cs, line 443 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-Kz5yHDrglQgIXas5bPs:-Kz5yHDrglQgIXas5bPt:b-idt2r4) ([raw file](https://github.com/sillsdev/l10nsharp/blob/65b951fa620ebf79e8324e92d7243565977d0137/src/ExtractXliff/Program.cs#L443)):*
> ```Smalltalk
> 							baseDoc.File.ProductVersion, _fileProductVersion);
> 					}
> 				}
> ```

Seems to me it's hardly worth a warning if baseDoc's product version is different from newDoc and _fileProductVersion...that just means (hopefully) we're updating to a new version of the product. But if newDoc's version is different from _fileProductVersion, then the command line arguments are telling us to lie about what version of the program this xliff matches...THAT deserves a warning. (Actually I'm wondering why we even have a command line argument for this...when would we NOT want to write the correct product version determined from the assemblies we processed into the new xliff?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-:-Kz6-KbaSoRRXu5eg8pI:b-w14d1x)*

StephenMcConnel:
Review status: 6 of 10 files reviewed at latest revision, 7 unresolved discussions.

---

*[src/ExtractXliff/Program.cs, line 255 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-KywL3rX7P5YhI7hFwqu:-Kz65tckvvbBNkrSZps0:b-qt9439) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0670eaf8de5bfd15a4f7184eb4f0e50a6005eb97/src/ExtractXliff/Program.cs#L255)):*

I still don't think it's OK that "--", which has a specific purpose of terminating the arguments list, will instead be taken as matching any argument not already looked for (since e.g., "--namespace".startsWith("--") is true)
</blockquote></details>

I added a check for "--" in the two places where it might be checked.

---

*[src/ExtractXliff/Program.cs, line 443 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-Kz5yHDrglQgIXas5bPs:-Kz661-RCyBc3uwZwmRo:b-2fw7kd) ([raw file](https://github.com/sillsdev/l10nsharp/blob/65b951fa620ebf79e8324e92d7243565977d0137/src/ExtractXliff/Program.cs#L443)):*

Seems to me it's hardly worth a warning if baseDoc's product version is different from newDoc and _fileProductVersion...that just means (hopefully) we're updating to a new version of the product. But if newDoc's version is different from _fileProductVersion, then the command line arguments are telling us to lie about what version of the program this xliff matches...THAT deserves a warning. (Actually I'm wondering why we even have a command line argument for this...when would we NOT want to write the correct product version determined from the assemblies we processed into the new xliff?)
</blockquote></details>

One reason for the file element attribute options is to preserve existing values so that files can be updated and show only real differences in strings to localize.  Changing the product -version does have some affect in crowdin as I recall, so it's not clear we want to change it in the xliff even when the program changes it.  Although ideally it would stay in sync with the program all else being equal.

And I'm inclined to leave the check in.  Without the warning, a file diff could show up and a developer decide to change the checked-in xliff file without being aware of the ramifications on crowdin of changing some of these attribute values.  (See DistFiles/localization/README.md for gory details.)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26)*

JohnThomson:
:lgtm:

---

Reviewed 3 of 4 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 5 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/26#-:-Kz9NYRw22sXFTc2uP1G:bnfp4nl)*

PR #28, opened 2017-11-28
Handle languages the system doesn't know about (BL-5260)

StephenMcConnel:
We can't depend on CultureInfo working for every language that our
programs may be localized into.  Mono 4 throws exceptions while
.Net 4.6.1 creates generally worthless objects for languages that
the authors didn't know about or didn't consider worth implementing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/28)

### DIFF
--- a/src/L10NSharp/L10NCultureInfo.cs
+++ b/src/L10NSharp/L10NCultureInfo.cs
@@ -5,54 +5,207 @@ using System.Linq;
 namespace L10NSharp
 {
 	/// <summary>
-	/// This class exists only to override Microsoft's NativeName for the Azeri neutral culture.
-	/// According to Ken Keyes the NativeName on the 'az-Latn' culture is correct, while the
-	/// NativeName on the 'az' culture is incorrect. So we do a swifty swap.
+	/// This class exists only to overcome the way CultureInfo is implemented to handle
+	/// only those cultures known to the authors of the .Net or Mono runtime.  On Linux,
+	/// trying to create a CultureInfo for an unknown culture throws an exception.  On
+	/// Windows for .Net 4.6, an empty object is created that cannot be modified.  This
+	/// class tries to create a CultureInfo object, and uses its values (for the most
+	/// part) if successful.  If it fails, well, there are a few languages we know
+	/// about that have been used for localization that we can handle in addition.
 	/// </summary>
-	public class L10NCultureInfo: CultureInfo
+	/// <remarks>
+	/// L10NCultureInfo objects will not provide any useful information for languages
+	/// that are not known to Windows .Net or to Mono (typically those without 2-letter
+	/// codes) and for which this class does not have any embedded information.
+	/// Therefore, this class should only be used for situations where we know only
+	/// a limited set of languages will occur, such as Bloom's UI languages.
+	/// It should not be used where the name passed in will be an arbitrary vernacular
+	/// language code.  (It won't crash, but it won't provide any information.)
+	/// </remarks>
+	public class L10NCultureInfo
 	{
-		private string _nativeName = null;
-
 		public L10NCultureInfo(string name)
-			:base(name)
 		{
-			// The Windows .Net runtime returns 'Azərbaycan dili (Azərbaycan)' for az-Latn, and
-			// something totally different for az.
-			// The Mono runtime returns "azərbaycan" for both az and az-Latn.
-			// So we just set this to what we "know" is the right value for az.
-			if (name == "az")
-				_nativeName = "Azərbaycan dili";
-		}
-
-		public override string NativeName
-		{
-			get
+			try
 			{
-				return _nativeName ?? base.NativeName;
+				RawCultureInfo = CultureInfo.GetCultureInfo(name);
+			}
+			catch (CultureNotFoundException ex)
+			{
+				RawCultureInfo = null;
+			}
+			if (RawCultureInfo == null || RawCultureInfo.EnglishName.StartsWith("Unknown Language"))
+			{
+				Name = name;
+				IsNeutralCulture = !Name.Contains("-");
+				if (IsNeutralCulture)
+				{
+					IetfLanguageTag = name;
+				}
+				else
+				{
+					var idx = name.IndexOf("-");
+					IetfLanguageTag = name.Substring(0, idx);
+				}
+				// CultureInfo returns 3-letter tags when a 2-letter tag doesn't exist.
+				// If it's a minor enough language to be unknown to .Net or Mono, it's
+				// not worth trying to find a 2-letter tag that usually wouldn't exist.
+				TwoLetterISOLanguageName = IetfLanguageTag;
+				ThreeLetterISOLanguageName = IetfLanguageTag;
+				switch (name)
+				{
+				case "pbu":
+					EnglishName = "Northern Pashto";
+					DisplayName = "Northern Pashto";
+					NativeName = "پښتو";
+					NumberFormat = CultureInfo.GetCultureInfo("ar").NumberFormat;
+					break;
+				case "prs":
+					EnglishName = "Dari";
+					DisplayName = "Dari";
+					NativeName = "دری";
+					NumberFormat = CultureInfo.GetCultureInfo("ar").NumberFormat;
+					break;
+				case "tpi":
+					EnglishName = "New Guinea Pidgin English";
+					DisplayName = "Tok Pisin";
+					NativeName = "Tok Pisin";
+					NumberFormat = CultureInfo.GetCultureInfo("en").NumberFormat;
+					break;
+				default:
+					EnglishName = string.Format("Unknown Language ({0})", name);
+					DisplayName = EnglishName;
+					NativeName = EnglishName;
+					NumberFormat = CultureInfo.InvariantCulture.NumberFormat;
+					break;
+				}
+			}
+			else
+			{
+				// The Windows .Net runtime returns 'Azərbaycan dili (Azərbaycan)' for az-Latn, and
+				// something totally different for az.
+				// The Mono runtime returns "azərbaycan" for both az and az-Latn.
+				// So we just set this to what we "know" is the right value for az.
+				if (name == "az")
+					NativeName = "Azərbaycan dili";
+				// The Windows .Net runtime returns 'Indonesia' for id.  It should be 'Bahasa Indonesia'.
+				else if (name == "id")
+					NativeName = "Bahasa Indonesia";
+				else
+					NativeName = RawCultureInfo.NativeName;
+				EnglishName = RawCultureInfo.EnglishName;
+				DisplayName = RawCultureInfo.DisplayName;
+				IsNeutralCulture = RawCultureInfo.IsNeutralCulture;
+				NumberFormat = RawCultureInfo.NumberFormat;
+				Name = RawCultureInfo.Name;
+				TwoLetterISOLanguageName = RawCultureInfo.TwoLetterISOLanguageName;
+				ThreeLetterISOLanguageName = RawCultureInfo.ThreeLetterISOLanguageName;
+				IetfLanguageTag = RawCultureInfo.IetfLanguageTag;
 			}
 		}
+
+		/// <summary>
+		/// Provide access to the underlying CultureInfo object if it exists.
+		/// </summary>
+		public CultureInfo RawCultureInfo { get; private set; }
+
+		// The following properties mimic those provided by CultureInfo.
+
+		public string NativeName { get; private set; }
+
+		public string DisplayName { get; private set; }
+
+		public string EnglishName { get; private set; }
+
+		public bool IsNeutralCulture { get; private set; }
+
+		public string Name { get; private set; }
+
+		public string TwoLetterISOLanguageName { get; private set; }
+
+		public string ThreeLetterISOLanguageName { get; private set; }
+
+		public string IetfLanguageTag { get; private set; }
+
+		public NumberFormatInfo NumberFormat { get; set; }
+
+		public static L10NCultureInfo CurrentCulture { get; internal set; }
 
 		/// <summary>
 		/// Gets the list of supported cultures in the form of L10NCultureInfo objects.
 		/// There is some danger in calling this repeatedly in that it creates new objects,
 		/// whereas the CultureInfo version appears to return cached objects.
 		/// </summary>
-		/// <param name="types"></param>
-		/// <returns></returns>
-		public new static IEnumerable<L10NCultureInfo> GetCultures(CultureTypes types)
+		public static IEnumerable<L10NCultureInfo> GetCultures(CultureTypes types)
 		{
-			return CultureInfo.GetCultures(types).Select(culture => new L10NCultureInfo(culture.Name));
+			var list = CultureInfo.GetCultures(types).Select(culture => new L10NCultureInfo(culture.Name)).ToList();
+			if ((types & CultureTypes.NeutralCultures) == CultureTypes.NeutralCultures)
+			{
+				bool havePbu = false;
+				bool havePrs = false;
+				bool haveTpi = false;
+				foreach (var ci in list)
+				{
+					if (ci.Name == "pbu")
+						havePbu = true;
+					else if (ci.Name == "prs")
+						havePrs = true;
+					else if (ci.Name == "tpi")
+						haveTpi = true;
+				}
+				if (!havePbu)
+					list.Add(GetCultureInfo("pbu"));
+				if (!havePrs)
+					list.Add(GetCultureInfo("prs"));
+				if (!haveTpi)
+					list.Add(GetCultureInfo("tpi"));
+			}
+			return list;
 		}
 
 		/// <summary>
 		/// Retrieves a new instance of a culture by using the specified culture name.
 		/// The CultureInfo version of this method returns a cached read-only instance.
 		/// </summary>
-		/// <param name="culture"></param>
-		/// <returns></returns>
-		public new static L10NCultureInfo GetCultureInfo(string culture)
+		public static L10NCultureInfo GetCultureInfo(string culture)
 		{
 			return new L10NCultureInfo(culture);
+		}
+
+		public override bool Equals(object obj)
+		{
+			var that = obj as L10NCultureInfo;
+			if (ReferenceEquals(that, null))
+				return false;
+			return (that.Name == this.Name) &&
+				(that.EnglishName == this.EnglishName);
+		}
+
+		public override int GetHashCode()
+		{
+			return Name.GetHashCode() + EnglishName.GetHashCode();
+		}
+
+		public override string ToString()
+		{
+			return string.Format("[L10NCultureInfo: Name={0}, EnglishName={1}]", Name, EnglishName);
+		}
+
+		public static bool operator ==(L10NCultureInfo ci1, L10NCultureInfo ci2)
+		{
+			if (ReferenceEquals(ci1, ci2))
+				return true;
+			if (ReferenceEquals(ci1, null))
+				return false;
+			if (ReferenceEquals(ci2, null))
+				return false;
+			return ci1.Equals(ci2);
+		}
+
+		// this is second one '!='
+		public static bool operator !=(L10NCultureInfo ci1, L10NCultureInfo ci2)
+		{
+			return !(ci1 == ci2);
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


JohnThomson:
Looks mostly good, with a few minor points.
There are classes in LibPalaso which know at least English names for all the Ethnologue languages. It's probably not desirable for L10NSharp to depend on LibPalaso, especially since SIL.Windows.Forms.WritingSystems already depends on L10NSharp; but I wonder whether there's any benefit in allowing a client to override the algorithm for naming unknown cultures? Tricky because ideally we'd want L10NSharp to define an interface and libpalaso to implement it, or slightly more simply for libpalaso to both define and implement it. Probably best left YAGNI.

---

Review status: 0 of 5 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 14 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-1yvhyyzA7lBhHj1GO:-L-1yvhyyzA7lBhHj1GP:b-u32brn) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L14)):*
> ```Smalltalk
> 	/// class tries to create a CultureInfo object, and uses its values (for the most
> 	/// part) if successful.  If it fails, well, there are a few languages we know
> 	/// about that have been used for localization that we can handle in addition.
> ```

Suggest adding something like, "L10NCultureInfo objects for languages that are not known to Windows or Mono (typically those without 2-letter codes) and for which this class does not supply information will not produce very useful L10NCultureInfo objects. Therefore, this class should only be used for situations where we know only a limited set of languages will occur, such as a list of Bloom's UI languages. It should not be used where the name passed in will be an arbitrary vernacular language code"

---

*[src/L10NSharp/L10NCultureInfo.cs, line 28 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-1ze_Cgt6_Bq1cV0yM:-L-1ze_Cgt6_Bq1cV0yN:buk02q0) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L28)):*
> ```Smalltalk
> 				RawCultureInfo = null;
> 			}
> 			if (RawCultureInfo == null || RawCultureInfo.EnglishName == "Unknown Language")
> ```

== or StartsWith? Seems like I've seen reference to things like Unknown Language (tpi).
Did you ever find out for sure whether "Unknown Language" will get localized in non-English versions of Windows?

---

*[src/L10NSharp/L10NCultureInfo.cs, line 35 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-2-5e6of21TmW_F159:-L-2-5e6of21TmW_F15A:bo4ti4k) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L35)):*
> ```Smalltalk
> 				{
> 					IetfLanguageTag = name;
> 					TwoLetterISOLanguageName = name;
> ```

Is this right even if it is NOT a two-letter code? I suppose it's better than any available alternative.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 43 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-2-LoFWVPmNYJTcb0R:-L-2-LoFWVPmNYJTcb0S:bc8ywe5) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L43)):*
> ```Smalltalk
> 					IetfLanguageTag = name.Substring(0, idx);
> 					TwoLetterISOLanguageName = IetfLanguageTag;
> 					ThreeLetterISOLanguageName = IetfLanguageTag;
> ```

Rather trivial, but those last two lines would work for the other case, too.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 71 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-20VIg6H9LWT9RGIUF:-L-20VIg6H9LWT9RGIUG:bbjg5mi) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L71)):*
> ```Smalltalk
> 					NumberFormat = CultureInfo.InvariantCulture.NumberFormat;
> 					break;
> 				}
> ```

Should we initialize NumberFormat to something in the case where we don't know anything?

---

*[src/L10NSharp/L10NCultureInfo.cs, line 87 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-207ywSanua9WLRAVR:-L-207ywSanua9WLRAVS:b-7xcbo6) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L87)):*
> ```Smalltalk
> 					NativeName = RawCultureInfo.NativeName;
> 				EnglishName = RawCultureInfo.EnglishName;
> 				DisplayName = RawCultureInfo.DisplayName;
> ```

Are there cases where any of these are null/empty and should be defaulted to one of the others?

---

*[src/L10NSharp/L10NCultureInfo.cs, line 147 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-21CBJVf425KehjCx1:-L-21CBJVf425KehjCx2:bul1yz5) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L147)):*
> ```Smalltalk
> 				}
> 				if (!havePbu)
> 					list.Add(GetCultureInfo("pbu"));
> ```

Seems unnecessarily complex. Why not put this line in the if above and drop the variable?

---

*[src/L10NSharp/L10NCultureInfo.cs, line 165 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-21PfKehZSQ-43CAqL:-L-21PfKehZSQ-43CAqM:b-gc72al) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L165)):*
> ```Smalltalk
> 		}
> 
> 		public override bool Equals (object obj)
> ```

.net coding standard says that if you override Equals, you MUST override GetHashCode(); it must return the same value for objects that Equals says are equal.
Do we need to do something so that == and != will behave correctly?

---

*[src/L10NSharp/LocalizationManager.cs, line 330 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-22-4ddxhqYvrcj_PG:-L-22-4ddxhqYvrcj_PH:bkcubq1) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/LocalizationManager.cs#L330)):*
> ```Smalltalk
> 			var ci = L10NCultureInfo.GetCultureInfo(langId);
> 			if (ci.RawCultureInfo != null)
> 				Thread.CurrentThread.CurrentUICulture = ci.RawCultureInfo;
> ```

Is there ANYTHING useful we can do about the thread culture in the other case? Seems it would be better to do something arbitrary but at least predictable...English or the Invariant culture...rather than leaving it set to whatever it was before.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-:-L-232AKBdZoHNUKyTpU:b26yne9)*

StephenMcConnel:
Review status: 0 of 5 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 14 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-1yvhyyzA7lBhHj1GO:-L-34wbmZNYgezls-S_w:blctv2l) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L14)):*

Suggest adding something like, "L10NCultureInfo objects for languages that are not known to Windows or Mono (typically those without 2-letter codes) and for which this class does not supply information will not produce very useful L10NCultureInfo objects. Therefore, this class should only be used for situations where we know only a limited set of languages will occur, such as a list of Bloom's UI languages. It should not be used where the name passed in will be an arbitrary vernacular language code"
</blockquote></details>

Done.  I edited your paragraph a bit and put it in a remarks section.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 28 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-1ze_Cgt6_Bq1cV0yM:-L-35BOwCcoQ9OKYKASG:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L28)):*

== or StartsWith? Seems like I've seen reference to things like Unknown Language (tpi).
Did you ever find out for sure whether "Unknown Language" will get localized in non-English versions of Windows?
</blockquote></details>

Done.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 35 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-2-5e6of21TmW_F159:-L-35wKG_OF2K98YkuQW:b-rlb3e0) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L35)):*

Is this right even if it is NOT a two-letter code? I suppose it's better than any available alternative.
</blockquote></details>

I added a commented explaining this.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 43 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-2-LoFWVPmNYJTcb0R:-L-35uJ16wRzFzo1Jb3j:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L43)):*

Rather trivial, but those last two lines would work for the other case, too.
</blockquote></details>

Done.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 71 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-20VIg6H9LWT9RGIUF:-L-366_lOifab84JvrbA:bsbavnq) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L71)):*

Should we initialize NumberFormat to something in the case where we don't know anything?
</blockquote></details>

Just in case someone blindly calls this with a truly unknown language and then expects number formatting to work.  I'd rather a user/tester complain that a new localization displayed as "Unknown Language" than complain that it did that and also crashed creating PDFs or whatever it is that uses number formatting.  THe InvariantCulture.NumberFormat seemed like a safe choice to prevent crashing at least.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 87 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-207ywSanua9WLRAVR:-L-36gow-0TZmfDqBXCz:b-fncqlh) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L87)):*

Are there cases where any of these are null/empty and should be defaulted to one of the others?
</blockquote></details>

I don't know of any such cases.  In Mono, CultureInfo.DisplayName always explicitly returns EnglishName.  I'm not sure how .Net works, but I would be surprised if it (or Mono) returned null for any culture that it claims to know about for any of these names.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 147 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-21CBJVf425KehjCx1:-L-37TmBYpkNtLYsIbWh:byrfovi) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L147)):*

Seems unnecessarily complex. Why not put this line in the if above and drop the variable?
</blockquote></details>

I think you're missing the ! (not) operator.  I have to scan the entire list to verify that the item does not exist before adding it.  I suppose I could create each of the objects and add them only if they were not already in the list.  But this approach avoids adding a (tiny) amount of wasted effort for when the system actually knows one or more of the languages we're adding if needed.  (I think Windows .Net does know about Dari.  I don't know if Mono is aware of any of the three languages we're adding.)

---

*[src/L10NSharp/L10NCultureInfo.cs, line 165 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-21PfKehZSQ-43CAqL:-L-39ORtnMqeohMSi5d7:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L165)):*

.net coding standard says that if you override Equals, you MUST override GetHashCode(); it must return the same value for objects that Equals says are equal.
Do we need to do something so that == and != will behave correctly?
</blockquote></details>

Done.

---

*[src/L10NSharp/LocalizationManager.cs, line 330 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-22-4ddxhqYvrcj_PG:-L-3NGHbhY_TfW_JE3Ji:b-1kx6qz) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/LocalizationManager.cs#L330)):*

Is there ANYTHING useful we can do about the thread culture in the other case? Seems it would be better to do something arbitrary but at least predictable...English or the Invariant culture...rather than leaving it set to whatever it was before.
</blockquote></details>

I think this setting affects only system dialogs.  Perhaps you're right.  I'll try InvariantCulture and we'll see if anyone ever complains.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28)*

JohnThomson:
Couple more questions...feel free to merge as-is if you don't think they're worth doing anything about.

---

Reviewed 3 of 5 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 71 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-20VIg6H9LWT9RGIUF:-L-3RxdWMUch73CcWLMT:b-x94rf4) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L71)):*

Just in case someone blindly calls this with a truly unknown language and then expects number formatting to work.  I'd rather a user/tester complain that a new localization displayed as "Unknown Language" than complain that it did that and also crashed creating PDFs or whatever it is that uses number formatting.  THe InvariantCulture.NumberFormat seemed like a safe choice to prevent crashing at least.
</blockquote></details>

Sorry, I was trying to get you to do what you'd already done. Somehow I missed that there is a default case.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 147 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-21CBJVf425KehjCx1:-L-3SyZyInU6HSIOHc2w:b1mvbcm) ([raw file](https://github.com/sillsdev/l10nsharp/blob/3f9a7f3de0096e9c227d68c316b2a744ad78a25a/src/L10NSharp/L10NCultureInfo.cs#L147)):*

I think you're missing the ! (not) operator.  I have to scan the entire list to verify that the item does not exist before adding it.  I suppose I could create each of the objects and add them only if they were not already in the list.  But this approach avoids adding a (tiny) amount of wasted effort for when the system actually knows one or more of the languages we're adding if needed.  (I think Windows .Net does know about Dari.  I don't know if Mono is aware of any of the three languages we're adding.)
</blockquote></details>

Yes, somehow I was mis-reading what you have. Sorry.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 76 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-3SMhbrVrilPCW_efs:-L-3SMhbrVrilPCW_eft:b-cisbcd) ([raw file](https://github.com/sillsdev/l10nsharp/blob/834aa4f6bae7cde7868cab51b300320c496077fd/src/L10NSharp/L10NCultureInfo.cs#L76)):*
> ```Smalltalk
> 					break;
> 				default:
> 					NativeName = "Unknown Language";
> ```

Missed this before, sorry...but should we make them all something like "Unknown Langauge (tpi)? Or even just use the code without emphasizing that it's unknown?

---

*[src/L10NSharp/L10NCultureInfo.cs, line 182 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-3YByu5noVSEC7u7Xr:-L-3YByu5noVSEC7u7Xs:b-6roi44) ([raw file](https://github.com/sillsdev/l10nsharp/blob/834aa4f6bae7cde7868cab51b300320c496077fd/src/L10NSharp/L10NCultureInfo.cs#L182)):*
> ```Smalltalk
> 			return (that.Name == this.Name) &&
> 				(that.EnglishName == this.EnglishName);
> 		}
> ```

I still think you should provide an implementation of == and !=. I find it very counter-intuitive that new L10NCultureInfo("fr").Equals(new L10NCultureInfo("fr") is true, while new L10NCultureInfo("fr") == (new L10NCultureInfo("fr") is false. Things I find on the web are ambivalent...MS apparently prefers not to mess with ==, but acknowledges it is appropriate in some cases, particularly immutable objects, which I believe yours is. Others agree with me. It's up to you...I won't insist, but if you don't do it I think a warning in the class comment might be appropriate.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-:-L-3Z7uruR-R1QtLflxU:bolr64x)*

StephenMcConnel:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/L10NSharp/L10NCultureInfo.cs, line 76 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-3SMhbrVrilPCW_efs:-L-3eLUlXq2Z-sABjUIF:b-z8kjgf) ([raw file](https://github.com/sillsdev/l10nsharp/blob/834aa4f6bae7cde7868cab51b300320c496077fd/src/L10NSharp/L10NCultureInfo.cs#L76)):*

Missed this before, sorry...but should we make them all something like "Unknown Langauge (tpi)? Or even just use the code without emphasizing that it's unknown?
</blockquote></details>

I made them all "Unknown Language (cod)".

---

*[src/L10NSharp/L10NCultureInfo.cs, line 182 at r2](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-L-3YByu5noVSEC7u7Xr:-L-3g7nH97TlbHjGItLw:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/834aa4f6bae7cde7868cab51b300320c496077fd/src/L10NSharp/L10NCultureInfo.cs#L182)):*

I still think you should provide an implementation of == and !=. I find it very counter-intuitive that new L10NCultureInfo("fr").Equals(new L10NCultureInfo("fr") is true, while new L10NCultureInfo("fr") == (new L10NCultureInfo("fr") is false. Things I find on the web are ambivalent...MS apparently prefers not to mess with ==, but acknowledges it is appropriate in some cases, particularly immutable objects, which I believe yours is. Others agree with me. It's up to you...I won't insist, but if you don't do it I think a warning in the class comment might be appropriate.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28)*

JohnThomson:
:lgtm:

---

Reviewed 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/28#-:-L-3jju-UwEqCIBD9IYO:bnfp4nl)*

PR #33, opened 2017-12-14
Add IgnoreExistingEnglishXliffFiles global flag

StephenMcConnel:
Also move XLiffDocument merging code to LocalizationManager.  The new
code helps in harvesting strings from running a program to verify which
strings in existing xliff files may be outdated and which strings need
to be added or updated.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/33)

### DIFF
--- a/src/ExtractXliff/ExtractXliff.csproj
+++ b/src/ExtractXliff/ExtractXliff.csproj
@@ -24,7 +24,7 @@
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
     <Externalconsole>true</Externalconsole>
-    <Commandlineparameters>-v -n Bloom -o Bloom.exe -p 3.1.000.0 -x /tmp/Bloom.xlf -b /d/github/BloomV4.0/DistFiles/localization/en/Bloom.xlf /d/github/BloomV4.0/output/Debug/Bloom.exe</Commandlineparameters>
+    <Commandlineparameters>-v -n Bloom -o Bloom.exe -p 3.1.000.0 -x /tmp/Bloom.xlf -b /d/github/BloomMaster/DistFiles/localization/en/Bloom.xlf /d/github/BloomMaster/output/Debug/Bloom.exe</Commandlineparameters>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <PlatformTarget>AnyCPU</PlatformTarget>
--- a/src/ExtractXliff/Program.cs
+++ b/src/ExtractXliff/Program.cs
@@ -83,151 +83,19 @@ namespace ExtractXliff
 			var baseDoc = LoadBaselineAndCompare(newDoc);
 
 			// Save the results to the output file, merging in data from the baseline XLIFF if one was specified.
-			MergeAndSaveXliffDataToFile(newDoc, baseDoc);
-		}
-
-		/// <summary>
-		/// Saves the merged XLIFF data to the output file.
-		/// </summary>
-		private static void MergeAndSaveXliffDataToFile(XLiffDocument xliffNew, XLiffDocument xliffOld)
-		{
-			// xliffNew has the data found in the current scan.
-			// xliffOld is that data from the (optional) input baseline XLIFF file.
-			// xliffOutput is the data that is actually written to the designated output XLIFF file.  It combines
-			//   data from both xliffNew and xliffOld.
-
-			// write the header elements of the new XLIFF file.
-			var xliffOutput = new XLiffDocument();
+			var xliffOutput = LocalizationManager.MergeXliffDocuments(newDoc, baseDoc, _verbose);
 			xliffOutput.File.SourceLang = kDefaultLangId;
 			if (!String.IsNullOrEmpty(_fileProductVersion))
 				xliffOutput.File.ProductVersion = _fileProductVersion;
 			else
-				xliffOutput.File.ProductVersion = xliffNew.File.ProductVersion;
+				xliffOutput.File.ProductVersion = newDoc.File.ProductVersion;
 			xliffOutput.File.HardLineBreakReplacement = kDefaultNewlineReplacement;
 			xliffOutput.File.AmpersandReplacement = kDefaultAmpersandReplacement;
 			xliffOutput.File.Original = _fileOriginal;
 			if (!String.IsNullOrEmpty(_fileDatatype))
 				xliffOutput.File.DataType = _fileDatatype;
 			else
-				xliffOutput.File.DataType = xliffNew.File.DataType;
-
-			var newStringCount = 0;
-			var changedStringCount = 0;
-			var wrongDynamicFlagCount = 0;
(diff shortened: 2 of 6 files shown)


JohnThomson:
I gather most of this is just moving code to a new location. Is there anything in particular that changed that I should look at? I don't think it's necessary to re-review code that just moved, so I'm happy for you just to merge unless you're doubtful about anything.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33#-:-L0LKtvfLibSGODtZOTE:b3gx7k2)*

StephenMcConnel:
The new code is mostly that in LocalizationManager.cs which revolves around IgnoreExistingEnglishXliffFiles.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33#-:-L0MMZmW0pccb6WL-qmz:b-j019ci)*

JohnThomson:
OK, I think I've had a reasonable look at the new stuff. A few minor and optional suggestions...up to you whether to act on any of them or just merge.

---

Reviewed 6 of 6 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/L10NSharp/LocalizationManager.cs, line 285 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33#-L0PaeTjLq79uGBQnxaF:-L0PaeTjLq79uGBQnxaG:bytn9tt) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c5012c32a560590aedc6060305c50e58604e4d1f/src/L10NSharp/LocalizationManager.cs#L285)):*
> ```Smalltalk
> 			if (!DefaultStringFileExistsAndHasContents() && File.Exists(defaultStringFileInstalledPath))
> 			{
> 				if (UILanguageId != "en" || !IgnoreExistingEnglishXliffFiles)
> ```

This expression or its negation occurs a few times and might be worth capturing in a method, especially since I think it has a quite specific purpose...something like "ScanningForCurrentStrings", right? That is, if the language is English and we're ignoring EnglishXliffFiles, we're using chunks of this code in a special way, not to provide localization of strings but to find what needs localizing.

---

*[src/L10NSharp/LocalizationManager.cs, line 1782 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33#-L0PbxXxI2zHLX_ZihRe:-L0PbxXyxbR5D01A1Stu:baou2g) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c5012c32a560590aedc6060305c50e58604e4d1f/src/L10NSharp/LocalizationManager.cs#L1782)):*
> ```Smalltalk
> 		/// directory.
> 		/// </summary>
> 		public static void MergeExistingEnglishXliffFileIntoNew(string installedStringFileFolder, string appId)
> ```

I'm assuming this long block of code is moved and not substantially changed.

---

*[src/L10NSharpTests/LocalizationManagerTests.cs, line 646 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33#-L0PcU0m-e-y86_Q_sY8:-L0PcU0m-e-y86_Q_sY9:b-et01pu) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c5012c32a560590aedc6060305c50e58604e4d1f/src/L10NSharpTests/LocalizationManagerTests.cs#L646)):*
> ```Smalltalk
> 
> 			var tu = mergedDoc.GetTransUnitForId("This.test");
> 			Assert.IsNotNull(tu);
> ```

I won't insist on changing this, but for future reference...how do you feel about the new(?) Assert.That? I find it much clearer to write something like Assert.That(tu.Source.Lang, Is.EqualTo("en")), though it's a fraction longer.
I'd also encourage you to look for opportunities to make helper assert methods that can make things shorter. For example, with the right helper method, a good deal of this could look something like
AssertExpectedTu(id="This.test", sourceLang="en", notesCount=2, secondNote="This is only a test", dynamic=true). (The first note is predictable, right, so you wouldn't need an argument for it?) Or you could have an array of strings called extraNotes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33#-:-L0PefF_Mp6USZ2yfbvi:bouhuis)*

StephenMcConnel:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/L10NSharp/LocalizationManager.cs, line 285 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33#-L0PaeTjLq79uGBQnxaF:-L0Q1gIqOiwuKYj2mtv5:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c5012c32a560590aedc6060305c50e58604e4d1f/src/L10NSharp/LocalizationManager.cs#L285)):*

This expression or its negation occurs a few times and might be worth capturing in a method, especially since I think it has a quite specific purpose...something like "ScanningForCurrentStrings", right? That is, if the language is English and we're ignoring EnglishXliffFiles, we're using chunks of this code in a special way, not to provide localization of strings but to find what needs localizing.
</blockquote></details>

Done.

---

*[src/L10NSharp/LocalizationManager.cs, line 1782 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33#-L0PbxXxI2zHLX_ZihRe:-L0QDkDvxQ--ri8Ox9Iw:b-nyimoq) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c5012c32a560590aedc6060305c50e58604e4d1f/src/L10NSharp/LocalizationManager.cs#L1782)):*

I'm assuming this long block of code is moved and not substantially changed.
</blockquote></details>

The changes are mostly cosmetic changes in the notes added about various states encountered.  Notably, I added a note today to record the old string value whenever the source string changes.

---

*[src/L10NSharpTests/LocalizationManagerTests.cs, line 646 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33#-L0PcU0m-e-y86_Q_sY8:-L0Q8P8jbj8-SLk6kaKf:b-sux1w5) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c5012c32a560590aedc6060305c50e58604e4d1f/src/L10NSharpTests/LocalizationManagerTests.cs#L646)):*

I won't insist on changing this, but for future reference...how do you feel about the new(?) Assert.That? I find it much clearer to write something like Assert.That(tu.Source.Lang, Is.EqualTo("en")), though it's a fraction longer.
I'd also encourage you to look for opportunities to make helper assert methods that can make things shorter. For example, with the right helper method, a good deal of this could look something like
AssertExpectedTu(id="This.test", sourceLang="en", notesCount=2, secondNote="This is only a test", dynamic=true). (The first note is predictable, right, so you wouldn't need an argument for it?) Or you could have an array of strings called extraNotes.
</blockquote></details>

Done.  I also added a utility method for creating a test TransUnit object.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/33)*

PR #41, opened 2018-01-24
Add flag to return only approved translations (BL-5472)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/41)

### DIFF
--- a/src/L10NSharp/LocalizationManager.cs
+++ b/src/L10NSharp/LocalizationManager.cs
@@ -47,6 +47,12 @@ namespace L10NSharp
 		/// </summary>
 		public static bool IgnoreExistingEnglishXliffFiles;
 
+		/// <summary>
+		/// Ignore any translated strings that are not marked "approved", acting as though the
+		/// translation didn't exist.
+		/// </summary>
+		public static bool ReturnOnlyApprovedStrings;
+
 		private static string s_uiLangId;
 		private static List<string> s_fallbackLanguageIds = new List<string>(new[] { kDefaultLang });
 
--- a/src/L10NSharp/XLiffUtils/XLiffDocument.cs
+++ b/src/L10NSharp/XLiffUtils/XLiffDocument.cs
@@ -143,7 +143,9 @@ namespace L10NSharp.XLiffUtils
 				{
 					Console.WriteLine("WARNING: string ID \"{0}\" already found in \"{1}\".", tu.Id, xLiffFile);
 				}
-				else
+				else if ((langId == xLiffDoc.File.SourceLang && langId == LocalizationManager.kDefaultLang) ||
+					!LocalizationManager.ReturnOnlyApprovedStrings ||
+					(tu.TranslationStatus == TransUnit.Status.Approved))
 				{
 					var target = tu.GetVariantForLang(langId);
 					if (target != null && !String.IsNullOrEmpty(target.Value))
--- a/src/L10NSharpTests/XLiffUtilsTests.cs
+++ b/src/L10NSharpTests/XLiffUtilsTests.cs
@@ -61,6 +61,90 @@ namespace L10NSharp.Tests
 			Assert.AreEqual(2, esdoc.NumberApproved);
 			Assert.AreEqual(3, esdoc.NumberTranslated);
 		}
+
+		[Test]
+		public void TestReturningAllStrings()
+		{
+			var enfile = Path.Combine(_testFolder, "Test.en.xlf");
+			var endoc = XLiffDocument.Read(enfile);
+			// These numbers may be counter-intuitive, but then the English isn't translated, is it?
+			// Code at the LocalizationManager level will make English look okay for display by
+			// faking the approved and translated counts.
+			Assert.AreEqual(4, endoc.StringCount);
+			string source;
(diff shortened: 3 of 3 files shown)


andrew-polk:
Reviewed 3 of 3 files at r1.

---

*[src/L10NSharpTests/XLiffUtilsTests.cs, line 106 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41#-L3hWTarp9TXZXpKjE_L:-L3hWTarp9TXZXpKjE_M:b6gioag) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e503d5adba6fbaaadad63da1051b7e15b5a11910/src/L10NSharpTests/XLiffUtilsTests.cs#L106)):*
> ```Smalltalk
> 			Assert.AreEqual("Ocultar el botón que abre la configuración.", source);
> 			Assert.IsTrue(esdoc.File.Body.TranslationsById.TryGetValue("SettingsProtection.PasswordDialog.FactoryPassword", out source));
> 			Assert.AreEqual("Factory Password", source);	// note it's stored and returned even though marked as needing translation
> ```

Is the desirable or de facto behavior? I find it strange that unapproved strings (with flag set) return false but untranslated strings return true with English. I would expect the same behavior in both cases.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41)*

StephenMcConnel:

---

*[src/L10NSharpTests/XLiffUtilsTests.cs, line 106 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41#-L3hWTarp9TXZXpKjE_L:-L3hp7_OdC88I8xVh7Ec:b4nphaj) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e503d5adba6fbaaadad63da1051b7e15b5a11910/src/L10NSharpTests/XLiffUtilsTests.cs#L106)):*

Is the desirable or de facto behavior? I find it strange that unapproved strings (with flag set) return false but untranslated strings return true with English. I would expect the same behavior in both cases.
</blockquote></details>

If you look at the actual file, the English string exists as the Spanish "translation" with an attribute indicating that it needs translation.  This is how Crowdin behaves, so I copied it in the test data.  In general, getting English back when no translation has been done is the de facto way that we've been using L10NSharp.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41)*

andrew-polk:

---

*[src/L10NSharpTests/XLiffUtilsTests.cs, line 106 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41#-L3hWTarp9TXZXpKjE_L:-L3hw3boQ6Jsjx8QRgE8:bh96y9v) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e503d5adba6fbaaadad63da1051b7e15b5a11910/src/L10NSharpTests/XLiffUtilsTests.cs#L106)):*

If you look at the actual file, the English string exists as the Spanish "translation" with an attribute indicating that it needs translation.  This is how Crowdin behaves, so I copied it in the test data.  In general, getting English back when no translation has been done is the de facto way that we've been using L10NSharp.
</blockquote></details>

Right. I was trying to figure out why we want this new behavior to be different or if we do.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41)*

StephenMcConnel:

---

*[src/L10NSharpTests/XLiffUtilsTests.cs, line 106 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41#-L3hWTarp9TXZXpKjE_L:-L3i9UhvF3bHqADWxZ21:b-fh071j) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e503d5adba6fbaaadad63da1051b7e15b5a11910/src/L10NSharpTests/XLiffUtilsTests.cs#L106)):*

Right. I was trying to figure out why we want this new behavior to be different or if we do.
</blockquote></details>

Strings in English as such come in from source instead of target, and are always implicitly both "translated" and "approved", even though explicitly neither flag would be set.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41)*

andrew-polk:

---

*[src/L10NSharpTests/XLiffUtilsTests.cs, line 106 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41#-L3hWTarp9TXZXpKjE_L:-L3iDNN8MDyeYoAFfOi3:btoz5km) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e503d5adba6fbaaadad63da1051b7e15b5a11910/src/L10NSharpTests/XLiffUtilsTests.cs#L106)):*

Strings in English as such come in from source instead of target, and are always implicitly both "translated" and "approved", even though explicitly neither flag would be set.
</blockquote></details>

Since the end result is always going to be the same (displaying English), I am not going to keep harping on this. 

I was just pointing out that from a consistency perspective, a caller to TryGetValue would expect to get the same result regardless of the reason for which a translation is unavailable.

The other reason this isn't worth belaboring is that the "right" thing to do is probably to change the original implementation such that TryGetValue returns false if it would return English regardless of the reason. But that is beyond the scope of this change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/41)*

PR #42, opened 2018-02-01
Fix a few more broken {[0-9]} marker patterns

StephenMcConnel:
These are really strange because there's no explanation for why
random LTR MARKER characters have been inserted into some strings.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/42)


andrew-polk:
:lgtm:

---

Reviewed 1 of 1 files at r1.

---

*[src/L10NSharp/LocalizedStringCache.cs, line 853 at r1](https://reviewable.io:443/reviews/sillsdev/l10nsharp/42#-L4Lh75FBYiMjLNzxGf6:-L4Lh75FBYiMjLNzxGf7:b1jpn5f) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a8d0d6a514c56df6a72caf0aa22648adde5dd1e4/src/L10NSharp/LocalizedStringCache.cs#L853)):*
> ```Smalltalk
> 								.Replace("{\u09EC}", "{6}").Replace("{\u09ED}", "{7}").Replace("{\u09EE}", "{8}")
> 								.Replace("{\u09EF}", "{9}");
> 			return target12;
> ```

For next time...
It would make it easier on devs to just use "target" all the way through rather than have a separate var name for each change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/l10nsharp/42#-:-L4LhRPTV6dBW1sKfsrg:bnfp4nl)*

PR #45, opened 2018-02-22
Handle null (or empty) language ids better (BL-5479)

StephenMcConnel:
This fixes a crashing unit test in Bloom.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/45)


StephenMcConnel:
Andrew already did a fix that was quite similar.  Not quite identical, but almost.

PR #64, opened 2019-06-11
Make TranslationManager.ReturnOnlyApprovedStrings behave dynamically

StephenMcConnel:
This supports optionally viewing unapproved translations as requested by
https://issues.bloomlibrary.org/youtrack/issue/BL-7281.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/64)


hatton:
I think we're going to hold off on this for now. We can ship what we need with a restart when the user enables this unapproved-messages flag.

StephenMcConnel:
This is obviously defunct.  The approach taken was not approved, any many changes have happened to the L10nSharp code.

Issue #66, opened 2019-07-05
Don't ask Bing translator to translate language names


tombogle:
In LanguageChoosingDialog.Application_Idle, it uses a BingTranslator to try to localize the apology message. Unfortunately, given the string:
**Our apologies, this program has not yet been localized for German (Deutsch). Please choose from one of the following languages:**
Bing translates _Deutsch_ as _English_. The string submitted to Bing should be the original format string, rather than the string with the parameters substituted in.

tombogle:
https://github.com/sillsdev/l10nsharp/pull/67

PR #67, opened 2019-07-05
Avoid having Bing mis-translate native UI language name back as "English"

tombogle:
Fix for https://github.com/sillsdev/l10nsharp/issues/66

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/67)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
-### Fixed
-- Don't ask Bing translator to translate language names: https://github.com/sillsdev/l10nsharp/issues/66.
-  Also don't display the name a second time in parentheses if English and native name are identical.
-
 ### Added
 
 - create symbol nuget package
@@ -27,6 +23,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - Find TMX files in `Generated` and `User Modified` directories
+
+- Don't ask Bing translator to translate language names: https://github.com/sillsdev/l10nsharp/issues/66.
+  Also don't display the name a second time in parentheses if English and native name are identical.
 
 ## [4.0.0] - 2019-05-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- Don't ask Bing translator to translate language names: https://github.com/sillsdev/l10nsharp/issues/66.
+  Also don't display the name a second time in parentheses if English and native name are identical.
+
 ### Added
 
 - create symbol nuget package
--- a/src/L10NSharp/L10NSharp-Designer.csproj
+++ b/src/L10NSharp/L10NSharp-Designer.csproj
@@ -123,6 +123,7 @@
     <Compile Include="UI\LanguageChoosingDialog.Designer.cs">
       <DependentUpon>LanguageChoosingDialog.cs</DependentUpon>
     </Compile>
+    <Compile Include="UI\LanguageChoosingDialogViewModel.cs" />
     <Compile Include="UI\LanguageChoosingSimpleDialog.cs">
       <SubType>Form</SubType>
     </Compile>
--- a/src/L10NSharp/UI/LanguageChoosingDialog.cs
+++ b/src/L10NSharp/UI/LanguageChoosingDialog.cs
@@ -1,6 +1,5 @@
 using System;
 using System.Drawing;
-using System.Globalization;
 using System.Windows.Forms;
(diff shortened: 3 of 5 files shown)


tombogle:
I further enhanced the fix. I could wrap some of that logic in a testable method. Obviously, we wouldn't want the tests to actually call Bing, which is where the real issues could lie. But the tests could at least try to anticipate some of the garbage that Bing could possibly throw back at us. So that might be worthwhile.

tombogle:
___
*[src/L10NSharp/UI/LanguageChoosingDialog.cs, line 31 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/67#-Lj1phjv4p_yyAZgRtks:-Lj2BNIoDnR9_zb6x9dQ:b2c4sep) ([raw file](https://github.com/sillsdev/l10nsharp/blob/71d474905d0ec8bfd2eb2c50558b17a4fd802a03/src/L10NSharp/UI/LanguageChoosingDialog.cs#L31)):*

Based on the old and following code, it appears that we expect _originalMessageTemplate to contain something like {0} ({1}), where (in English) {0} is the English name of the language and {1} is the native name, like German (Deutsch). We want to give Bing German ({1}) and get back Deutsch ({1}) and THEN change {1} to "German".
I would expect this to fail because the string.Format() will choke on a string containing {1} when only given a single argument. You'd need to pass "{1}" as a second argument (or I think you could pass "{0}" as the second argument, then Bing will get German ({0}), hopefully return Deutsch ({0}, and you are all ready to insert the English).
Even if I'm wrong and string.Format gives back German ({1}) and Bing translates this to Deutsch ({1}), I think this needs more explanation. The trailing comment
on the line above, particularly, was just confusing...
I think it would be good to document what it's acceptable for _originalMessageTemplate to contain, what we want to pass to Bing for each such case, what we hope to get back, and then be clearer about happy and unhappy paths for what we do next and how we tell the difference.
</blockquote></details>

I did consider doing the replacement as part of the initial call to Format. (I've done that before in other apps, but it kind of feels confusing either way.) I'm doing some refactoring to bring this code under unit tests. I'll think it through and try to make it intelligible.

tombogle:
___
*[src/L10NSharp/UI/LanguageChoosingDialog.cs, line 31 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/67#-Lj1phjv4p_yyAZgRtks:-Lj2aiVLAVMHJ9epzdSc:b-i1c7ad) ([raw file](https://github.com/sillsdev/l10nsharp/blob/71d474905d0ec8bfd2eb2c50558b17a4fd802a03/src/L10NSharp/UI/LanguageChoosingDialog.cs#L31)):*

I did consider doing the replacement as part of the initial call to Format. (I've done that before in other apps, but it kind of feels confusing either way.) I'm doing some refactoring to bring this code under unit tests. I'll think it through and try to make it intelligible.
</blockquote></details>

And, yes, you were right.

tombogle:
___
*[src/L10NSharp/UI/LanguageChoosingDialog.cs, line 31 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/67#-Lj1phjv4p_yyAZgRtks:-Lj2iu-i6u4qKbiER_V3:b-pue765) ([raw file](https://github.com/sillsdev/l10nsharp/blob/71d474905d0ec8bfd2eb2c50558b17a4fd802a03/src/L10NSharp/UI/LanguageChoosingDialog.cs#L31)):*

Catch (Exception) is dangerous!
</blockquote></details>

Maybe so. That is existing code (in a new file). It's probably overkill, but since the only purpose of this code is to attempt to replace the English version of the message with a language-specific one, any unhandled error is only going to make things worse. Since it's calling out to a third-party web-based delegate to get the translation it might be difficult to anticipate everything that could possibly go wrong. I had actually considered whether it might be better to display the translated message in addition to the original English one, just in case we get something back that is garbage (and also because if someone needs help, it could be hard for a support person to make sense of the problem. Searching for the translated string in our code and TMX files won't result in a hit.) In the end, though, I figured that could be left for a rainier day.

ermshiperete:
___
*[src/L10NSharp/UI/LanguageChoosingDialogViewModel.cs, line 18 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/67#-LjFIlpE5BMg0tXneiRA:-LjFIlpE5BMg0tXneiRB:b64pbm6) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5f1146bec1b2c2a487adc3d3f217cf9771d2dd3d/src/L10NSharp/UI/LanguageChoosingDialogViewModel.cs#L18)):*
> ```Smalltalk
> 			_messageLabelFormat = messageLabelFormat;
> 			AcceptButtonText = _acceptButtonText = acceptButtonText;
> 			WindowTitle = _windowTitle = windowTitle;
> ```

What's the purpose of having an auto property `WindowTitle` (and `AcceptButtonText`) and a separate field `_windowTitle`? Wouldn't the auto property be sufficient?

PR #68, opened 2019-07-08
In LanguageChoosingDialog, handle off-line case where translator just returns source string

tombogle:
If translator returns an unmodified source string, don't substitute t…he English language name for the vernacular name.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/68)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+
+- If translator returns an unmodified source string, don't substitute the English language name for the vernacular name.
 
 ## [4.0.1] - 2019-07-08
 
--- a/src/L10NSharp/UI/LanguageChoosingDialog.cs
+++ b/src/L10NSharp/UI/LanguageChoosingDialog.cs
@@ -20,7 +20,7 @@ namespace L10NSharp.UI
 		void Application_Idle(object sender, EventArgs e)
 		{
 			Application.Idle -= Application_Idle;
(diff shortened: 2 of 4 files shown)


tombogle:
___
*[src/L10NSharp/UI/LanguageChoosingDialogViewModel.cs, line 30 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/68#-LjI_KTq-XTWC_SKJPg-:-LjMF7r4EPScw-U5nf42:b3go4k3) ([raw file](https://github.com/sillsdev/l10nsharp/blob/aefb86603a36158d295eefcbaf40f2b87e6f9cb7/src/L10NSharp/UI/LanguageChoosingDialogViewModel.cs#L30)):*

I think this method is now correct, given a string that contains placeholders for up to two arguments and some unwritten assumptions about their being language names and how we want those handled and what the translator will do with language names. The method name baffles me. It doesn't, in any sense I can think of, save a translator for later use ("set which translator to use"), nor does it work to take 'Set' as a noun. It's more like SetMessageUsingTranslator. I think it would be very helpful to give some examples of expected input strings, what we hope Bing will do, and the expected end result. However, those problems were pre-existing, and the PR is already an improvement. So it's up to you whether to improve it more or just merge.
</blockquote></details>

Since it actually (in the common case) modifies the message as well as the accept button text and window title, I renamed it to TranslateStrings. I added summary comments to the constructor that should satisfy your concern about the format string. The existing tests hopefully provide adequate examples of the kinds of things the translator might do (though I didn't make one that attempts to mimic Bing verbatim).

PR #74, opened 2020-04-23
Add instructions for building with GitVersion

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/74)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,16 @@ The command line build command would look something like this:
 Note that on Linux building L10NSharp requires at least version 5 of Mono, which `mono5-sil` provides.
 The Mono 5 version that the Mono project provides also works (if mono is at least version 5.16).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,6 +49,16 @@ The command line build command would look something like this:
 Note that on Linux building L10NSharp requires at least version 5 of Mono, which `mono5-sil` provides.
 The Mono 5 version that the Mono project provides also works (if mono is at least version 5.16).
 
+To include the correct version number requires [GitVersion](https://gitversion.net/docs/usage/command-line), which can be installed using
```
ermshiperete, line 52: I don't think that's true. Yes, we're using GitVersion for versioning, but the .csproj files that require it reference GitVersionTask which takes care of generating the version numbers.

You'd only have to install the GitVersion command line tool if you'd want to see what version GitVersion would produce, or to test out settings or the like.

PR #75, opened 2020-05-26
Added version of LocalizationManager.Create with param to allow additional GetString methods

tombogle:
This allows the client to specify additional localization methods when extracting from code. Useful, for example, when using SIL.Localizer to allow L10nSharp to be injected in code that does not depend on it.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/75)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - Add `ExtractXliff` tool as nuget package
 - Add `CheckOrFixXliff` tool as nuget package
+- Added version of LocalizationManager.Create to allow "custom" localization methods
 
 ## [4.0.3] - 2020-01-21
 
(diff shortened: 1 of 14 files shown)


tombogle:
___
*[src/L10NSharp/LocalizationManager.cs, line 132 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/75#-M8GrMPIAtq20FvW-EG4:-M8H1gMY73zgYQSNxYq3:b-rik1jy) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5fc31bd73618512f4b1c79cc023e068fc2502ecf/src/L10NSharp/LocalizationManager.cs#L132)):*

> ```
> 		/// additional methods that should be regarded as calls to get localizations. If the method
> 		/// is named "Localize", its signature will be parsed as if it were an extension method
> 		/// where the first parameter is the English string and the second parameter is the ID.
> 		/// Otherwise, it will be treated like a GetString method where the first parameter is
> 		/// the ID and the second parameter is the English string. In both cases, the following
> 		/// additional optional parameters are permitted in this order: comment, tootip, shortcut
> ```
Seems worth adding some more of these as test cases...
For example, the `Localize` method name, tooltip, shortcut keys.
</blockquote></details>

Probably. Though all the descibed behavior is based on underlying code that didn't actually change.

PR #76, opened 2020-05-28
PG-1365: Enhanced ExtractXliff tool to be able to specify additional methods

tombogle:
This allows us to harvest strings from non Windows.Forms assemblies in libpalaso (that can't reference L10nSharp) but that use the SIL.Localizer class as a proxy.

### DIFF
--- a/src/ExtractXliff/Program.cs
+++ b/src/ExtractXliff/Program.cs
@@ -3,7 +3,9 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Reflection;
+using System.Text.RegularExpressions;
 using L10NSharp;
 using L10NSharp.CodeReader;
 using L10NSharp.XLiffUtils;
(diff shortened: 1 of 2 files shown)


src/ExtractXliff/Program.cs
```diff
@@ -187,9 +222,33 @@ static bool AddNamespace(string[] args, ref int i)
...
+		static bool AddLocalizationMethod(string[] args, ref int i)
+		{
+			// -m  --method = additional localization method [zero or more]
+			++i;
+			if (i < args.Length)
+			{
+				if (args[i] == "--")
+					return false;
+
+				var match = Regex.Match(args[i], @"^(?<namespace>[a-zA-Z_][a-zA-Z0-9_]*(\.[a-zA-Z_][a-zA-Z0-9_]*)*)\.(?<class>[a-zA-Z_][a-zA-Z0-9_]*)\.(?<methodName>[a-zA-Z_][a-zA-Z0-9_]*)$");
```
jasonleenaylor, line 234: Future developers will at a quick glance say "Oh of course, use the regular expression to capture the content following the last '.' to method name, the second to the last '.' as class and the content before that as namespace. "
Or maybe they will get a headache.
Maybe a brief comment here or on the method would help

src/ExtractXliff/Program.cs
```diff
@@ -71,16 +76,40 @@ static void Main(string[] args)
...
+							.GetMethods(BindingFlags.Static | BindingFlags.Public)
+							.Where(m => m.Name == methodNameSpec.Item3));
+
+						if (_verbose)
+							Console.WriteLine($"Method {methodNameSpec.Item2}.{methodNameSpec.Item3} in {asm.GetName().FullName} will be treated as a localization method.");
+						_additionalLocalizationMethodNames.RemoveAt(index--);
+					}
+					catch
+					{
+					}
```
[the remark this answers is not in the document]
tombogle: Done (and improved even)

src/ExtractXliff/Program.cs
tombogle: Betterized

PR #78, opened 2020-06-05
Create LocalizationManager using more specific locale if necessary

tombogle:
For programs that are switching from TMX-based localization to XLIFF-based, languages like Spanish (es) and Portuguese (pt) will typically use a langiage code that includes a country specifier (es-ES, pt-PT). The LocalizationManager already knows about these mappings and handles them ocrrectly when getting strings, but not when checking to see if the requested UI language is available.
If the caller requests a desired UI based on a two-letter language code that is not available, but there is a mapping to the corresponding (country-)specifiic locale, use that without displaying the "UI not available" dialog.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/78)

### DIFF
--- a/src/L10NSharp/LocalizationManagerInternal.cs
+++ b/src/L10NSharp/LocalizationManagerInternal.cs
@@ -59,7 +59,7 @@ namespace L10NSharp
 				desiredUiLangId = CultureInfo.CurrentUICulture.TwoLetterISOLanguageName;
 			}
 
-			if (!DesiredUiCultureIsAvailable(desiredUiLangId))
+			if (!IsDesiredUiCultureAvailable(desiredUiLangId))
 			{
 				using (var dlg = new LanguageChoosingDialog(L10NCultureInfo.GetCultureInfo(desiredUiLangId), applicationIcon))
 				{
@@ -75,13 +75,16 @@ namespace L10NSharp
 			return lm;
 		}
 
-		private static bool DesiredUiCultureIsAvailable(string desiredUiLangId)
+		private static bool IsDesiredUiCultureAvailable(string desiredUiLangId)
 		{
 			var ci = L10NCultureInfo.GetCultureInfo(desiredUiLangId);
 			if (GetUILanguages(true).Contains(ci))
 				return true;
+			// If this fallback implementation seems strange, read the really long summary
+			// comment for the unit test:
+			// Create_PreferredUiLanguageIsGenericVariant_CreatesLocalizationManagerForSpecificVariant
 			return MapToExistingLanguage.TryGetValue(desiredUiLangId, out desiredUiLangId) &&
-				DesiredUiCultureIsAvailable(desiredUiLangId);
+				IsDesiredUiCultureAvailable(desiredUiLangId);
 		}
 
 		/// ------------------------------------------------------------------------------------
--- a/src/L10NSharp/LocalizationManagerInternal.cs
+++ b/src/L10NSharp/LocalizationManagerInternal.cs
@@ -59,10 +59,9 @@ namespace L10NSharp
 				desiredUiLangId = CultureInfo.CurrentUICulture.TwoLetterISOLanguageName;
 			}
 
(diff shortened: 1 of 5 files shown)


tombogle:
I initially made the tests generic, but the TMX version failed because TMX doesn't have these mappings. That might just be coincidence, based on the test data. I didn't try to track it down, because in practice, all I care about is getting the behavior when switching from TMX to XLIFF. I could add a comment to the test (and production code) to explain that. (Or I could try to figure out why it is that way, but I don't want to.)

tombogle:
___
*[src/L10NSharpTests/LocalizationManagerXliffTests.cs, line 13 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/78#-M950WaI0y-WEd-93fHI:-M954hHD-2Jf-C7qp2ub:bjd4892) ([raw file](https://github.com/sillsdev/l10nsharp/blob/cdffcbbc684d8e560c2b54ec3906d0df38588c87/src/L10NSharpTests/LocalizationManagerXliffTests.cs#L13)):*

> ```
> ensures that if the UI pops up, a developer will not have time to respond
> ```
3 seconds is a long time for a dev to respond (unless it is running on a TC agent, of course).
I think I would tweak this comment.
</blockquote></details>

It's 3 seconds total. I tested it. 3 seconds was defnitely not enough time to respond. I didn't want to make it too short, lest it fail spuriously on a bogged down machine.

tombogle:
___
*[src/L10NSharpTests/LocalizationManagerXliffTests.cs, line 17 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/78#-M950ndkCs2yXNC0WGc7:-M957YwW6mKPfY0AW7Cc:b-kgi2rg) ([raw file](https://github.com/sillsdev/l10nsharp/blob/cdffcbbc684d8e560c2b54ec3906d0df38588c87/src/L10NSharpTests/LocalizationManagerXliffTests.cs#L17)):*

> ```
> "es"
> ```
Don't we want two test cases here? One for "es" and one for "es-ES"?
</blockquote></details>

I don't think so. At least in practice, I don't think we'll have a case where it goes the other diirection. I guess I could write the test and see if it even works.

tombogle:
___
*[src/L10NSharpTests/LocalizationManagerXliffTests.cs, line 13 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/78#-M950WaI0y-WEd-93fHI:-M96fuwz1FbDLGoYIKGy:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/cdffcbbc684d8e560c2b54ec3906d0df38588c87/src/L10NSharpTests/LocalizationManagerXliffTests.cs#L13)):*

I think I would still tweak it.
Maybe something like:

If the test set up is incorrect, we can end up with a dialog popping up and asking the user which UI Language to use.
In that case, we don't want the test to hang, but rather to fail.
However, we also don't want the developer to interact with the dialog which could give a false success. 
So we set the time out to 3 seconds, which seems to be long enough to let the test run successfully but short enough to prevent a developer from interacting with it.
</blockquote></details>

Done.

tombogle:
___
*[src/L10NSharp/LocalizationManagerInternal.cs, line 85 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/78#-M9HYk-O5rMViFMIzwGQ:-M9IpkOY0ApZkq6JB32w:bszpry) ([raw file](https://github.com/sillsdev/l10nsharp/blob/b141f98d7e201709a6db4444615a608376487843/src/L10NSharp/LocalizationManagerInternal.cs#L85)):*

Hmm, makes me wonder: is the method name really long because the comment is really long, or is it the other way round? :wink: 
</blockquote></details>

Probably the two things are related. I've written worse. This test is just a bit funky, and I figured a long comment was needed to make it make sense. I couldn't think of any good way to shorten the name and still have it be fairly meaningful if someone ever sees it fail.

PR #87, opened 2021-08-19
Generate messages for all extraction errors

papeh:
If the result of extracting strings is null or the wrong type,
or if some other exception occurs, report it in the localizations
file. This gives developers information and prevents crashes later.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/87)


ermshiperete (approved):
lgtm

PR #93, opened 2022-04-04
Make lazy loading of xliffs thread-safe (BL-11086)

JohnThomson:
Also makes all methods consistently handle fall-back from e.g. es-BR to es-ES.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93)

### DIFF
--- a/src/L10NSharp/XLiffUtils/XLiffBody.cs
+++ b/src/L10NSharp/XLiffUtils/XLiffBody.cs
@@ -68,42 +68,14 @@ namespace L10NSharp.XLiffUtils
 		}
 
 		#region Properties
-		/// ------------------------------------------------------------------------------------
-		/// <summary>
-		/// Gets the list of translation units in the file.
-		/// Note, a new list is generated each time, so this can be inefficient. Consider using
-		/// TransUnitsUnordered if you don't absolutely need a list. This method is retained
-		/// mainly for backwards compatibility.
-		/// Modifying the list returned (unlike in previous versions) will have no effect on
-		/// the XliffBody.
-		/// The list returned is not in any predictable order, specifically NOT the same order
-		/// as any list previously passed to the setter.
-		/// </summary>
-		/// ------------------------------------------------------------------------------------
-		[XmlIgnore]
-		public List<XLiffTransUnit> TransUnits
-		{
-			get
-			{
-				var result = TransUnitsUnordered.ToList();
-				return result;
-			}
-			set
-			{
-				_transUnitDict.Clear();
-				foreach (var tu in value)
-				{
-					AddTransUnit(tu);
-				}
-			}
-		}
 
 		/// <summary>
 		/// This property exists solely to support serializing and deserializing an XliffBody in a
-		/// backwards-compatible way. Previously, we simply put this annotation on TransUnits.
-		/// However, the serialization code assumes it can get the list and use Add to put things
-		/// into it, which does not work with the current version. The List wrapper implements just
-		/// the functions required of an IEnumerable property that can be serialized and deserialized.
+		/// backwards-compatible way. The serialization code assumes it can get the object and use
+		/// Add to put things into it when deserializing as well as running the enumeration to get
+		/// the things to serialize. The ListWrapper implements just those necessary functions.
+		/// This property must be public for the serializer to work, but is not intended for use
+		/// by other clients.
 		/// </summary>
 		[XmlElement("trans-unit")]
 		public ListWrapper TransUnitsForXml
@@ -115,7 +87,6 @@ namespace L10NSharp.XLiffUtils
 				return new ListWrapper(result, this);
 			}
 		}
-
 
 		[XmlIgnore] public IEnumerable<XLiffTransUnit> TransUnitsUnordered => _transUnitDict.Values;
 
@@ -203,7 +174,7 @@ namespace L10NSharp.XLiffUtils
 			_transUnitDict[key] = tu;
 			return true;
 		}
-		internal bool AddTransUnit(XLiffTransUnit tu)
+		public bool AddTransUnit(XLiffTransUnit tu)
 		{
 			if (!AddTransUnitRaw(tu))
 				return false;
@@ -245,7 +216,7 @@ namespace L10NSharp.XLiffUtils
 		/// Removes the specified translation unit.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
-		internal void RemoveTransUnit(XLiffTransUnit tu)
+		public void RemoveTransUnit(XLiffTransUnit tu)
 		{
 			// if the ID is null, it can't be in our dictionary, unless someone
 			// cheated and changed the ID by putting it there after inserting it.
--- a/src/L10NSharp/XLiffUtils/XLiffBody.cs
+++ b/src/L10NSharp/XLiffUtils/XLiffBody.cs
@@ -14,8 +14,11 @@
 // </remarks>
 // ---------------------------------------------------------------------------------------------
 using System;
+using System.Collections;
+using System.Collections.Concurrent;
 using System.Linq;
 using System.Collections.Generic;
+using System.Threading;
 using System.Xml.Serialization;
 using L10NSharp.XLiffUtils;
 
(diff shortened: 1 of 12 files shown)


andrew-polk (changes requested):
LGTM (besides a minor existing issue I brought up), but I think another set of eyes would be beneficial.

> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-:-MzqGwCHAatsoQmXyomw:bythwps)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/L10NSharp/LocalizationManagerInternal.cs, line 96 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-MzpvkGF1EuWaBZyMzXa:-MzpvkGGA29djhedpiO9:b-f9uedd) ([raw file](https://github.com/sillsdev/l10nsharp/blob/12f13a957b95faa0dc3eb1de0ab3499dbc260c61/src/L10NSharp/LocalizationManagerInternal.cs#L96)):*
> ```Smalltalk
> 			var fallbackLangId = MapToExistingLanguageIfPossible(desiredUiLangId);
> 			if (fallbackLangId == desiredUiLangId)
> 				return false; // identity mappings don't give us new options, and could lead to stack overflow
> ```

I see you didn't change the logic here, but it is counterintuitive for something called `IsDesiredUiCultureAvailable` to return false just because we didn't have to map it to anything.
Is it actually true that we should never get here because we would have returned true three lives above? If so, we should probably add that to the comment.

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

___
*[src/L10NSharp/LocalizationManagerInternal.cs, line 96 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-MzpvkGF1EuWaBZyMzXa:-MzqiCrL76qaDj9FvhCu:bbza80c) ([raw file](https://github.com/sillsdev/l10nsharp/blob/12f13a957b95faa0dc3eb1de0ab3499dbc260c61/src/L10NSharp/LocalizationManagerInternal.cs#L96)):*

I see you didn't change the logic here, but it is counterintuitive for something called `IsDesiredUiCultureAvailable` to return false just because we didn't have to map it to anything.
Is it actually true that we should never get here because we would have returned true three lives above? If so, we should probably add that to the comment.
</blockquote></details>

I hope it's clear now.

andrew-polk (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @JohnThomson)

tombogle (commented):
LGTM. Is there a good way to write a unit test or two that can ensure that this is correct?

jasonleenaylor (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93)* status: all files reviewed, 1 unresolved discussion (waiting on @JohnThomson)

___
*[src/L10NSharp/LocalizationManagerInternal.cs, line 104 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-MzuhADLCFIih4QkJY5M:-MzuhADLCFIih4QkJY5N:b-3qmx4i) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c83df1fb3861883b4849a9da8086dbb46174155b/src/L10NSharp/LocalizationManagerInternal.cs#L104)):*
> ```Smalltalk
> 			// localization. In such a case, if we don't detect it, we'd be making a recursive call
> 			// with the same argument, which would produce stack overflow. So if we didn't find
> 			// a different language to try, we must fail now.
> ```

John, this comment very accurately captures the situation and is certainly more helpful to understanding than the original comment, but it is also more likely to generate a tldr; response.

Imagining myself maintaining this code I think only the details about our usage in this context is necessary. e.g. Why do we return false on this condition?

I'd sum this up as:
// If the input and output of MapToExistingLanguageIfPossible are the same then there is no mapping known for the language and we should return false instead of infinitely recursing.

Feel free to come up with your own summary, but try to limit your creative writing to two or three lines.



_Code quote:_
```Smalltalk
			// However, MapToExistingLanguageIfPossible often returns the code we started with.
			// This can indicate that we don't need to map it since we have exactly that localization,
			// but in such a case, IsLocalizationAvailable would have returned true above and we would
			// not get here. However, it also returns an unchanged code for a locale about which we
			// know nothing at all...there is no closely related language for which we have a
			// localization. In such a case, if we don't detect it, we'd be making a recursive call
			// with the same argument, which would produce stack overflow. So if we didn't find
			// a different language to try, we must fail now.
```
___
*[src/L10NSharp/LocalizationManagerInternal.cs, line 694 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-MzukPxzDn2b-hpJunf8:-MzukPxzDn2b-hpJunf9:bcpsh5u) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c83df1fb3861883b4849a9da8086dbb46174155b/src/L10NSharp/LocalizationManagerInternal.cs#L694)):*
> ```Smalltalk
> 
> 
> 				var pieces = langId.Split('-');
> ```

Move this after the comment to help clarify the comment.



_Code quote:_
```Smalltalk
var pieces = langId.Split('-');
```

ermshiperete (approved):

JohnThomson (commented):
I'm open to ideas. For the most part, I've made things concurrent by using .NET ConcurrentDictionary. Assuming that works correctly, there should not be much danger of race conditions; and I don't believe anything I've changed holds a lock for more than the duration of a single call, nor while calling back to unknown code, nor ever holds more than one lock at a time; so I don't think I've created a danger of deadlocks. I'm hoping the existing unit tests cover what I've changed pretty well

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-:-MzzyR3i2M8LxzytRYez:b2la29l)* status: 1 of 9 files reviewed, all discussions resolved (waiting on @andrew-polk and @jasonleenaylor)

___
*[src/L10NSharp/LocalizationManagerInternal.cs, line 104 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-MzuhADLCFIih4QkJY5M:-N---u0s7QpX6BOMWvfI:bb7f2mf) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c83df1fb3861883b4849a9da8086dbb46174155b/src/L10NSharp/LocalizationManagerInternal.cs#L104)):*
<details><summary><i>Previously, jasonleenaylor (Jason Naylor) wrote…</i></summary><blockquote>

John, this comment very accurately captures the situation and is certainly more helpful to understanding than the original comment, but it is also more likely to generate a tldr; response.

Imagining myself maintaining this code I think only the details about our usage in this context is necessary. e.g. Why do we return false on this condition?

I'd sum this up as:
// If the input and output of MapToExistingLanguageIfPossible are the same then there is no mapping known for the language and we should return false instead of infinitely recursing.

Feel free to come up with your own summary, but try to limit your creative writing to two or three lines.

</blockquote></details>

done
___
*[src/L10NSharp/LocalizationManagerInternal.cs, line 694 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-MzukPxzDn2b-hpJunf8:-N---qtVBU5sGDGfKBRH:bb7f2mf) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c83df1fb3861883b4849a9da8086dbb46174155b/src/L10NSharp/LocalizationManagerInternal.cs#L694)):*
<details><summary><i>Previously, jasonleenaylor (Jason Naylor) wrote…</i></summary><blockquote>

Move this after the comment to help clarify the comment.
</blockquote></details>

done

JohnThomson (commented):
I don't know how to write unit tests specifically for thread-safety.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-:-N--0T4s5buz0pYT4Kei:b-pn2620)* status: 1 of 9 files reviewed, all discussions resolved (waiting on @andrew-polk and @jasonleenaylor)

src/L10NSharp/XLiffUtils/XLiffBody.cs
```diff
@@ -29,35 +32,93 @@ public class XLiffBody
...
 				}
+			}
+		}
 
-				return _transUnits;
+		/// <summary>
+		/// This property exists solely to support serializing and deserializing an XliffBody in a
+		/// backwards-compatible way. Previously, we simply put this annotation on TransUnits.
+		/// However, the serialization code assumes it can get the list and use Add to put things
+		/// into it, which does not work with the current version. The List wrapper implements just
```
tombogle, line 105: ListWrapper

tombogle (changes requested):

src/L10NSharp/XLiffUtils/XLiffBody.cs
```diff
@@ -29,35 +32,93 @@ public class XLiffBody
...
+			}
+		}
+
 		#region Properties
 		/// ------------------------------------------------------------------------------------
 		/// <summary>
 		/// Gets the list of translation units in the file.
+		/// Note, a new list is generated each time, so this can be inefficient. Consider using
+		/// TransUnitsUnordered if you don't absolutely need a list. This method is retained
+		/// mainly for backwards compatibility.
```
tombogle, line 76: Since the behavior is subtly changing, maybe trying to preserve "backwards compatibility" is a bad thing.

ermshiperete (changes requested):
This PR deserves an entry in `CHANGELOG.md` :smile:

JohnThomson (commented):
Done

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-:-N-4cF459HA5fdVyu8bJ:bb6un1j)* status: 1 of 12 files reviewed, 2 unresolved discussions (waiting on @andrew-polk, @jasonleenaylor, and @tombogle)

___
*[src/L10NSharp/XLiffUtils/XLiffBody.cs, line 76 at r5](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-MzzweOukyNNXkbDWXkU-r5-76:-N-4dact1yx4Hyi1WGii:b-8qt6ii) ([raw file](https://github.com/sillsdev/l10nsharp/blob/86f91a710b4c6f15f1c080b31abee8e5eec8d9f0/src/L10NSharp/XLiffUtils/XLiffBody.cs#L76)):*

Since the behavior is subtly changing, maybe trying to preserve "backwards compatibility" is a bad thing.
</blockquote></details>

Following your suggestion in Slack, I decided to remove it completely, making it a clearly breaking change, though unlikely to break most clients.
___
*[src/L10NSharp/XLiffUtils/XLiffBody.cs, line 105 at r5](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-MzzweOukyNNXkbDWXkU-r5-105:-N-4dftv0lM2DqRk3_eY:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/86f91a710b4c6f15f1c080b31abee8e5eec8d9f0/src/L10NSharp/XLiffUtils/XLiffBody.cs#L105)):*

ListWrapper
</blockquote></details>

Done.

ermshiperete (approved):
A few nitpicks in `CHANGELOG.md`, but otherwise :lgtm:

JohnThomson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93)* status: 1 of 12 files reviewed, 5 unresolved discussions (waiting on @andrew-polk, @ermshiperete, @jasonleenaylor, and @tombogle)

___
*[CHANGELOG.md, line 25 at r7](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-N-4c6yrA3ZSSvKbhgsV-r7-25:-N-8HFrG35fV9vdVKWvH:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6be6ac18917df11bfd7b63919e577dc758030bb4/CHANGELOG.md#L25)):*

This should go in the (new) `### Removed` section
</blockquote></details>

Done.
___
*[CHANGELOG.md, line 27 at r7](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-N-4c6yrA3ZSSvKbhgsV-r7-27:-N-8HHHtD93lTsR9TVSv:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6be6ac18917df11bfd7b63919e577dc758030bb4/CHANGELOG.md#L27)):*

The three "added" topics should go in the `### Added` section
</blockquote></details>

Done.
___
*[CHANGELOG.md, line 30 at r7](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-N-4c6yrA3ZSSvKbhgsV-r7-30:-N-8HO2V3EmU__GtmmSD:b-p1343i) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6be6ac18917df11bfd7b63919e577dc758030bb4/CHANGELOG.md#L30)):*

```suggestion
  serialization and deserialization in XML, but is not intended for any other purpose.
```
</blockquote></details>

I 'm thinking you just wanted this indented?

jasonleenaylor (commented):
> Reviewed 1 of 7 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93)* status: 3 of 12 files reviewed, 5 unresolved discussions (waiting on @andrew-polk, @ermshiperete, @jasonleenaylor, and @tombogle)

___
*[src/L10NSharp/LocalizationManagerInternal.cs, line 104 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-MzuhADLCFIih4QkJY5M:-N--04a_ATtQHLzTMefX:b-8fx2u7) ([raw file](https://github.com/sillsdev/l10nsharp/blob/c83df1fb3861883b4849a9da8086dbb46174155b/src/L10NSharp/LocalizationManagerInternal.cs#L104)):*

done
</blockquote></details>

:+1:

ermshiperete (approved):

ermshiperete (approved):
:lgtm:

> Reviewed 4 of 7 files at r3, 2 of 3 files at r5, 1 of 3 files at r6, 1 of 1 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-:-N-CjXy83LQ2-z2nPbOG:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @tombogle)

___
*[CHANGELOG.md, line 30 at r7](https://reviewable.io/reviews/sillsdev/l10nsharp/93#-N-4c6yrA3ZSSvKbhgsV-r7-30:-N-CjR7IAa1r9muuVh5N:bb7sa0c) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6be6ac18917df11bfd7b63919e577dc758030bb4/CHANGELOG.md#L30)):*

I 'm thinking you just wanted this indented?
</blockquote></details>

yes.

PR #98, opened 2022-06-02
Fixed bug: LoadedManagers has wrong AppId because it was passed in with extension during Create

tombogle:
.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/98)

### DIFF
--- a/src/L10NSharp/LocalizationManagerInternal.cs
+++ b/src/L10NSharp/LocalizationManagerInternal.cs
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 SIL International
+// Copyright (c) 2022 SIL International
 // This software is licensed under the MIT License (http://opensource.org/licenses/MIT)
 
 using System;
@@ -199,10 +199,17 @@ namespace L10NSharp
 			IEnumerable<MethodInfo> additionalLocalizationMethods,
 			params string[] namespaceBeginnings)
 		{
+			if (string.IsNullOrWhiteSpace(appId))
+				throw new ArgumentNullException(nameof(appId));
+			var origExeExtension = Path.GetExtension(appId);
+			if (origExeExtension == string.Empty)
+				origExeExtension = ".dll";
+			appId = Path.GetFileNameWithoutExtension(appId);
+
 			return Create(desiredUiLangId, appId, appName,
 				relativeSettingPathForLocalizationFolder, applicationIcon,
 				directoryOfWritableXliffFiles =>
-					(ILocalizationManagerInternal<T>) new XLiffLocalizationManager(appId, appName,
+					(ILocalizationManagerInternal<T>) new XLiffLocalizationManager(appId, origExeExtension, appName,
 						appVersion, directoryOfInstalledXliffFiles,
 						directoryOfWritableXliffFiles, directoryOfWritableXliffFiles,
 						additionalLocalizationMethods,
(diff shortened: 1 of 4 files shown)


andrew-polk (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/98)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`src/L10NSharp/LocalizationManagerInternal.cs` line 62 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/98#-N3_EDzrDxkfcLBneZvf:-N3_EDzrDxkfcLBneZvg:bbaf0rq) ([raw file](https://github.com/sillsdev/l10nsharp/blob/f017f23ccc04ba90cd9e2b15bc2a33616695e529/src/L10NSharp/LocalizationManagerInternal.cs#L62)):*
> ```Smalltalk
> 			{
> 				lm = createMethod(directoryOfWritableTranslationFiles);
> 				appId = lm.Id;
> ```

I don't really understand what is going on here. Maybe at this point, we should be using a variable named something other than appId?
Or maybe just a comment is warranted?

But you say this is a blocker, so I'm ok merging as is if you need to.

tombogle:
*[`src/L10NSharp/LocalizationManagerInternal.cs` line 62 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/98#-N3_EDzrDxkfcLBneZvf:-N3_HMBjCfSzHvybTzwB:b-yjlyk5) ([raw file](https://github.com/sillsdev/l10nsharp/blob/f017f23ccc04ba90cd9e2b15bc2a33616695e529/src/L10NSharp/LocalizationManagerInternal.cs#L62)):*

I don't really understand what is going on here. Maybe at this point, we should be using a variable named something other than appId?
Or maybe just a comment is warranted?

But you say this is a blocker, so I'm ok merging as is if you need to.
</blockquote></details>

Actually, the code IS really confusing... and wrong. I think it would almost always end up working, but just because we use it in a way that doesn't fall into the hole.

andrew-polk (changes requested):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/98)* status: all files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`src/L10NSharp/LocalizationManagerInternal.cs` line 206 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/98#-N3_fXuN4dIlhcDwvF0T:-N3_fXuOAUtQQR7XFr9n:b-dk3iti) ([raw file](https://github.com/sillsdev/l10nsharp/blob/ce36777422869395d3640a45d0aff60af1c5bac6/src/L10NSharp/LocalizationManagerInternal.cs#L206)):*
> ```Smalltalk
> 			var origExeExtension = Path.GetExtension(appId);
> 			if (origExeExtension == string.Empty)
> 				origExeExtension = ".dll";
> ```

I may just be missing something, but Bloom's usage here is to pass "Bloom" for both appId and name. 
But it is Bloom.exe, not Bloom.dll. So it feels like something is wrong with defaulting to .dll.

Though I do see that is what the XLiffLocalizationManager constructor was defaulting to... so I guess I am missing something.

PR #100, opened 2022-07-07
Added optional owner parameter to methods that show dialog boxes

tombogle:
This causes them to be displayed centered on a parent window (and not appear off-screen).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/100)

### DIFF
--- a/src/L10NSharp/UI/TipDialog.cs
+++ b/src/L10NSharp/UI/TipDialog.cs
@@ -25,10 +25,20 @@ namespace L10NSharp.UI
 			set { _icon.Image = value; }
 		}
 
-		public static void Show(string message)
+		public static void ShowAltShiftClickTip(IWin32Window owner = null)
+		{
+			Show("If you click on an item while you hold alt and shift keys down, this tool will " +
+				"open up with that item already selected.", owner);
+		}
+
+		public static void Show(string message, IWin32Window owner)
 		{
 			using (var d = new TipDialog(message, "Tip", SystemIcons.Information.ToBitmap()))
-				d.ShowDialog();
+			{
+				if (owner != null)
+					d.StartPosition = FormStartPosition.CenterParent;
+				d.ShowDialog(owner);
+			}
 		}
 
 		private TipDialog()
--- a/src/L10NSharp/UI/TipDialog.cs
+++ b/src/L10NSharp/UI/TipDialog.cs
@@ -7,25 +7,19 @@ namespace L10NSharp.UI
 {
 	public partial class TipDialog : Form
 	{
-		public string DialogTitle
-		{
-			get { return Text; }
-			set { Text = value; }
-		}
-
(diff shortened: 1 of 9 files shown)


tombogle:
*[`src/L10NSharp/ILocalizationManager.cs` line 95 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/100#-N6PmPESAS6eXHoh2udp:-N6QN9eFEN8NcLKX3xGG:bvbozws) ([raw file](https://github.com/sillsdev/l10nsharp/blob/7ebe4e831b6a7f031e576fa44fb6f743011e1b02/src/L10NSharp/ILocalizationManager.cs#L95)):*

I don't know about this myself, but I thought you (or @ermshiperete?) told me that adding an optional parameter is still a breaking change. I thought I was told that you have to add an override to make it truly not breaking. But you can also just make it major change if you're trying to get this in before you publish 5.0 (since it will still just be 5.0).
</blockquote></details>

Yes, that's true. I wondered about it when I went to push the change today (which I had left ready to push for some reason), but I was foolishly too lazy check.

src/L10NSharp/UI/TipDialog.cs
```diff
@@ -25,10 +25,20 @@ public new Image Icon
 			set { _icon.Image = value; }
 		}
 
-		public static void Show(string message)
+		public static void ShowAltShiftClickTip(IWin32Window owner = null)
```
ermshiperete, line 28: Is this method expected to be called from clients, or should this be `internal` ?

tombogle:
*[`src/L10NSharp/UI/TipDialog.cs` line 28 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/100#-N6PUDs5jjo1b0XWIJEn-r2-28:-N6T8081D3jfjYVGUQqh:b-8miqs0) ([raw file](https://github.com/sillsdev/l10nsharp/blob/947c45cd7e8a6cdfddc6deab7683c8bb2c5cbdb0/src/L10NSharp/UI/TipDialog.cs#L28)):*

Is this method expected to be called from clients, or should this be `internal` ?
</blockquote></details>

I was thinking it should be public since the existing Show method was. I made this new method so as to avoid having the identical hard-coded string in multiple places. But now that I consider further, I guess it's probably not very likely  that an application would want to show this particular tip in any other context. In actual fact, based on a search in Github, it does not appear that any of our projects use the TipDialog to display other tips at all. So probably we could just make the whole class internal. But I guess it's harmless and it does seem to be intended for other uses.

andrew-polk (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/100)* status: all files reviewed, 1 unresolved discussion (waiting on @ermshiperete)

PR #102, opened 2022-08-23
Modernize build

ermshiperete:
- Add GitHub actions
- Use NUnit 3
- Update build dependencies
- Move Directory.Build.props to source root
- Allow to build with dotnet

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/102)


src/L10NSharpTests/LocalizationExtenderTests.cs
```diff
@@ -185,7 +190,8 @@ public void PrepareDataGridViewColumnsTests()
 		/// group for each.
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
-		[Test, Ignore("The test body was commented out before I got here, why run it?")]
+		[Test]
+		[Ignore("The test body was commented out before I got here, why run it?")]
```
papeh, line 194: If it's been commented out that long, maybe it should be deleted.

papeh (approved):

src/CheckOrFixXliff/Program.cs
```diff
@@ -130,7 +131,7 @@ private static void FixBrokenTransUnit(XElement tu, XmlNamespaceManager namespac
 			var target = tu.XPathSelectElement("x:target", namespaceManager);
 			if (source.HasElements || target.HasElements)
 			{
-				Console.WriteLine("Cannot fix {0} because the translated material contains XML elements", tuid);
+				Console.WriteLine(@"Cannot fix {0} because the translated material contains XML elements", tuid);
```
papeh, line 134: I don't think this `@` is changing anything (here and throughout).

src/ExtractXliff/Program.cs
```diff
@@ -91,26 +92,28 @@ static void Main(string[] args)
...
 						if (_verbose)
-							Console.WriteLine($"Method {methodNameSpec.Item2}.{methodNameSpec.Item3} in {asm.GetName().FullName} will be treated as a localization method.");
+							Console.WriteLine($@"Method {methodNameSpec.Item2}.{methodNameSpec.Item3} in {asm.GetName().FullName} will be treated as a localization method.");
 						_additionalLocalizationMethodNames.RemoveAt(index--);
 					}
 					catch (Exception e)
 					{
 						if (_verbose)
-							Console.WriteLine("Error using reflection on {asm.GetName().FullName} to get type {methodNameSpec.Item2} or method {methodNameSpec.Item3}:" + e.Message);
+							Console.WriteLine($@"Error using reflection on {asm.GetName().FullName} to get type {methodNameSpec.Item2} or method {methodNameSpec.Item3}:" + e.Message);
```
papeh, line 101: e.Message could be included in the interpolated string

src/CheckOrFixXliff/Program.cs
ermshiperete: It marks it as verbatim string and so R# knows that it's not a localizable string.

src/ExtractXliff/Program.cs
ermshiperete, line 101:
```suggestion
							Console.WriteLine($@"Error using reflection on {asm.GetName().FullName} to get type {methodNameSpec.Item2} or method {methodNameSpec.Item3}: {e.Message}");
```

src/CheckOrFixXliff/Program.cs
papeh: Thanks. That's what I thought, but the option to ignore verbatim strings had been turned off in FieldWorks, so I had trained myself to stop doing that.

papeh (approved):

PR #105, opened 2022-10-06
Replace exception rethrow with writing warning message (20221006)

StephenMcConnel:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/105)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,12 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Scanning resources for strings no longer rethrows unexpected exceptions.  It now writes the
+  exception (and stack trace) using a (conditional) Console.WriteLine and a Debug.WriteLine.
+  Rethrowing the exception leads to creating a zero-length xliff file which causes another
+  exception.  Swallowing the exception allows the scanning process to continue and complete.
+  The old behavior has been an endless source of periodic instability in using L10NSharp over
+  the years.
 - remove progress dialog when initializating Xliff localization managers (BL-11157)
 - Made string retrieval operations on Xliff-based LocalizationManagers thread-safe
 - Added ILocalizationManager parameter to StringsLocalizedHandler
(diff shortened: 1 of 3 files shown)


StephenMcConnel (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/105)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @tombogle)

___
*[`src/L10NSharp/CodeReader/StringExtractor.cs` line 115 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/105#-NDi41gPCkahAx4Z7DRC:-NDj4Zxp3NSP0_LNdVuP:b-wzdwtz) ([raw file](https://github.com/sillsdev/l10nsharp/blob/ddd6e384f6d4d3ef071f1ddbbfbfb383146f66d9/src/L10NSharp/CodeReader/StringExtractor.cs#L115)):*

I see the point of this, but I wonder if we have done enough to try to figure out what specific scenarios can cause this and analyzed them to determine if something more appropriate can be done? I see there are three non-test callers of this method. If some of them could perhaps benefit from getting back a list of errors that could be logged or reported in a more helpful way, that might be worth considering. An overload of this method could make that possible. If there is a known way to force one of these unexpected exceptions in a test, it might also be a good idea to write a test case that shows that the exception is eaten (or reported back, if we go that route).
</blockquote></details>

I've extended the fix to include storing a public list of the warning messages and reporting the number of ignored exceptions for each unit of extraction.  I don't see any easy way of creating a unit test with an artificial exception throw, and don't think it's worth the programmer time to figure out something.  This is fixing a real problem that is very sporadic but was causing my flatpak build of Bloom 5.3 to just not run at all.
We've had years to figure out scenarios that cause this, and I would guess the flatpak scenario is that the restricted sandbox environment lacks something obscure that is wanted by the mono runtime for deciphering some sort of date/time format that just doesn't matter.

ermshiperete (approved):

tombogle (approved):
> Reviewed 1 of 2 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/105)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @StephenMcConnel)

PR #107, opened 2022-11-29
Add `LocalizationManager.Create` methods without `kind` parameter

ermshiperete:
Also deprecate existing `Create` methods with `kind` parameter. 

Since `Xliff` is now the only supported translation memory, the `kind` parameter is unnecessary. This change adds new overloads without the `kind` parameter and marks the existing methods as `Obsolete`. 

+semver:minor

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/107)

### DIFF
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -34,7 +34,7 @@ jobs:
     - name: Install .NET Core
       uses: actions/setup-dotnet@c0d4ad69d8bd405d234f1c9166d383b7a4f69ed8 # v2.1.0
       with:
-        dotnet-version: 6.0.x
+        dotnet-version: 7.0.x
 
     - name: Build and Pack
       run: dotnet build --configuration Release
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -44,7 +44,8 @@ jobs:
       if: matrix.os == 'ubuntu-latest'
 
     - name: Test (Windows)
-      run: dotnet test --configuration Release --no-build -- NUnit.TestOutputXml=TestResults
+      # specify path to test assembly as workaround for https://github.com/nunit/nunit3-vs-adapter/issues/1040
+      run: dotnet test --configuration Release --no-build output\Release\net461\*Tests.dll -- NUnit.TestOutputXml=TestResults
       if: matrix.os != 'ubuntu-latest'
 
     - name: Upload Test Results
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- `LocalizationManager.Create` methods without `TranslationMemory kind` parameter
+
+### Deprecated
+
+- `LocalizationManager.Create` methods with `TranslationMemory kind` parameter
+
 ## [6.0.0] - 2022-11-21
 
 ### Changed
--- a/L10NSharp.sln
+++ b/L10NSharp.sln
@@ -20,6 +20,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SolutionFolder", "SolutionF
 		CHANGELOG.md = CHANGELOG.md
 		GitVersion.yml = GitVersion.yml
 		README.md = README.md
+		Directory.Build.props = Directory.Build.props
+		Directory.Build.targets = Directory.Build.targets
 	EndProjectSection
 EndProject
 Global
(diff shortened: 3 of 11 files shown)


andrew-polk (approved):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/107)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tombogle)

tombogle (approved):
I think this is okay. At JohnT's suggestion, I had left the scaffolding in place to allow for a future scenario where something other than XLIFF might be needed. But I think this is an improvement. You might want to consider running it by John to be sure he's satisfied.

> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/107#-:-NI33XGW6Pm-v3UJolz3:bsz6liy)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

JohnThomson (approved):
I have some questions and suggestions. But I'm OK with merging as-is, too.

I'm puzzled that the Test Results above says it removes 7 tests. I don't see any changes that would remove tests.

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/107#-:-NIidKe05fPMwHp-HZY6:beac8m3)* status: all files reviewed, 2 unresolved discussions (waiting on @ermshiperete)

___
*[`src/L10NSharp/LocalizationManager.cs` line 193 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/107#-NIiYMyP7RawkwxN6rDM:-NIiYMyP7RawkwxN6rDN:b-9y8l5x) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e7f1189f97a32d68130184a9fccdc6889ddb4bad/src/L10NSharp/LocalizationManager.cs#L193)):*
> ```Smalltalk
> 		/// files installed with the application.</param>
> 		/// <param name="relativeSettingPathForLocalizationFolder">The path, relative to
> 		/// %appdata%, where your application stores user settings (e.g., "SIL\SayMore").
> ```

Sure this is not %localappdata%? Most of our code seems to use that.
___
*[`src/L10NSharp/LocalizationManager.cs` line 201 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/107#-NIiYv357BK5oGqbFLK7:-NIiYv357BK5oGqbFLK8:b-6zf7wf) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e7f1189f97a32d68130184a9fccdc6889ddb4bad/src/L10NSharp/LocalizationManager.cs#L201)):*
> ```Smalltalk
> 		/// what types to scan for localized string calls. For example, to only scan
> 		/// types found in Pa.exe and assuming all types in that assembly begin with
> 		/// 'Pa', then this value would only contain the string 'Pa'.</param>
> ```

It's a pity we have to duplicate all of the above comments. For  a reader of the code, it would be much nicer to just say "Create an instance. This overload may be used where there are no additionalLocalizationMethods.

But then intellisense won't help with this overload. 

What would you think of making additionalLocalizationMethods default to null and having just the one overload? I suppose that might be slightly more work for anyone correcting old code that used the obsolete Create and had namespaceBeginnings but no additionalLocalizationMethods; but I suspect the common case is to have neither. Or perhaps it becomes confusing if you want to pass null for additionalLocalizationMethods, but that could be interpreted as the first namespaceBeginnings so the compiler will complain?

I also wonder if there's any good reason for namespaceBeginnings to be 'params'. That's useful for commonly used methods that take a single list, but here, I wonder whether it's worth the possible ambiguity that it introduces. It's not very difficult to create an array of strings, even if it contains only one thing, in just one place in the calling code. If we get rid of 'params', the API can just have two optional params, and they can't be confused because one is an enumerable of MethodInfos where the other is an array of strings.

tombogle:
There's a TestFixture that runs a set of tests for each type of storage. I'm guessing that removing the kind parameter affects how that test works.

JohnThomson (approved):
:lgtm:. I'll let someone else make the final call on merging, though I think all is now ready.

> Reviewed 4 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/107#-:-NIm7cQz5HvSonc0dC_t:b65w02e)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

andrew-polk (approved):
> Reviewed 4 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/107)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

tombogle (approved):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/107)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

Issue #109, opened 2023-03-03
Initializing without a country code should find l10ns with a country code


papeh:
* Install Palaso with palaso.zh-CN.xlf
* `var palasoLocMan = LocalizationManager.Create("zh", "Palaso", "Palaso", version, installedL10nBaseDir, userL10nBaseDir, CommonResources.chorus, FlexLocalizationEmailAddress, "SIL");`
Expected: a zh-CN LocalizationManager is initialized w/o user input
Actual: a dialog says Chinese is unavailable and prompts the user to choose a different language from { English, Chinese (China) }

This would resolve https://jira.sil.org/browse/LT-21232, but that was also resolved client-side.

papeh:
Do we care what happens if both zh-CN and zh-Hans (or -HK, -TW, or -AU) are available?
If a client requests zh-HK but only zh-CN is available, should we pick that? (presently, the answer is yes in another area of L10nSharp)
Barring any strong opinions, I plan to leave the above behaviour undefined.

tombogle:
I'm not an expert on Chinese, but these can be tricky questions. For some languages, including Chinese, there is more than one script, and while the spoken language might me more or less the same, there is no guarantee that a person who reads one script will have any ability to read another. Where there is a single predominant script, an generic language tag should probably select that script without asking. Where there are multiple scripts or country-specific variants with no clear default, it's presumably better to ask than to blindy select one. But if the differences between the two are relatively minor (e.g., en-US vs. en-GB), there might be times when it is better to select one and leave it to the user to select the preferred one later if they need to.

tombogle:
See https://www.w3.org/International/articles/bcp47/#macro

papeh:
Right, this is tricky. Unfortunately, L10nSharp doesn't work with language tags apart from extracting the bare language code to find matches. Without shipping langtags.json, it would be impossible to know whether we're looking at en-US vs. en-GB or zh-Hans vs. zh-Hant (simplified or traditional script). Do you think the existing code should be changed to require an exact or more generic match?

I agree that it may be better to allow the user to choose if there are multiple matches.

papeh:
[...]

PR #110, opened 2023-03-03
[API] #109: find specific languages when generics are specified

papeh:
If a client tries to initialize a LocalizationManager with only
a language code, and we have a language available with that code
plus extra information, use that language without prompting the user.
Previous behavior: prompted the user with a dialog in this case.
Unchanged behavior: initializing with extra information will fall back to the bare language code.

This addresses https://github.com/sillsdev/l10nsharp/issues/109
This addresses https://jira.sil.org/browse/LT-21232

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/110)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - `LocalizationManager.Create` methods without `TranslationMemory kind` parameter
 
+### Fixed
+
+- `LocalizationManager.Create("es"` loads `es-ES` if it is the best match
+
 ### Deprecated
 
 - `LocalizationManager.Create` methods with `TranslationMemory kind` parameter
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
-- `LocalizationManager.Create("es"` loads `es-ES` if it is the best match
+- `LocalizationManager.Create("es"` loads `es-ES` if it is the best match (previously, this resulted in a dialog making the user choose)
 
 ### Deprecated
 
--- a/L10NSharp.sln
+++ b/L10NSharp.sln
@@ -6,7 +6,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "L10NSharp", "src\L10NSharp\
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "L10NSharpTests", "src\L10NSharpTests\L10NSharpTests.csproj", "{BCE5B569-057C-4D7E-832E-E44A2DA705AC}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SampleApp", "src\SampleApp\SampleApp.csproj", "{58923B30-FD84-4BCC-85E0-607DCA7E7C95}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SampleApp", "src\SampleApp\SampleApp.csproj", "{58923B30-FD84-4BCC-85E0-607DCA7E7C95}"
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ExtractXliff", "src\ExtractXliff\ExtractXliff.csproj", "{E0666C78-B8DC-4232-952C-753940D54921}"
 EndProject
@@ -18,10 +18,10 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SolutionFolder", "SolutionF
 		.gitattributes = .gitattributes
 		.gitignore = .gitignore
 		CHANGELOG.md = CHANGELOG.md
-		GitVersion.yml = GitVersion.yml
-		README.md = README.md
 		Directory.Build.props = Directory.Build.props
 		Directory.Build.targets = Directory.Build.targets
+		GitVersion.yml = GitVersion.yml
+		README.md = README.md
 	EndProjectSection
 EndProject
 Global
@@ -54,6 +54,17 @@ Global
 		{BCE5B569-057C-4D7E-832E-E44A2DA705AC}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
 		{BCE5B569-057C-4D7E-832E-E44A2DA705AC}.Release|Mixed Platforms.Build.0 = Release|Any CPU
 		{BCE5B569-057C-4D7E-832E-E44A2DA705AC}.Release|x86.ActiveCfg = Release|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|Mixed Platforms.Build.0 = Debug|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|x86.ActiveCfg = Debug|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|x86.Build.0 = Debug|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|Any CPU.Build.0 = Release|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|Mixed Platforms.Build.0 = Release|Any CPU
+		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|x86.ActiveCfg = Release|Any CPU
 		{E0666C78-B8DC-4232-952C-753940D54921}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
 		{E0666C78-B8DC-4232-952C-753940D54921}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{E0666C78-B8DC-4232-952C-753940D54921}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
@@ -78,15 +89,6 @@ Global
 		{E4BB984D-DFB8-42EF-860D-5A038FF85B1C}.Release|Mixed Platforms.Build.0 = Release|Any CPU
 		{E4BB984D-DFB8-42EF-860D-5A038FF85B1C}.Release|x86.ActiveCfg = Release|Any CPU
 		{E4BB984D-DFB8-42EF-860D-5A038FF85B1C}.Release|x86.Build.0 = Release|Any CPU
-		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
-		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|Any CPU.Build.0 = Debug|Any CPU
-		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|Mixed Platforms.ActiveCfg = Debug|Any CPU
-		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Debug|Mixed Platforms.Build.0 = Debug|Any CPU
-		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|Any CPU.ActiveCfg = Release|Any CPU
-		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|Any CPU.Build.0 = Release|Any CPU
-		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
-		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|Mixed Platforms.Build.0 = Release|Any CPU
-		{58923B30-FD84-4BCC-85E0-607DCA7E7C95}.Release|x86.ActiveCfg = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ or `es/Whatever.xlf`).
 ### Selecting a Language
 
 If an exact match for the requested language is not available, L10NSharp will try to find the best available language. For example, if the client
-requests `es` but only `Whatever.es-ES.xlf` is available, `Whatever.es-ES.xlf` will be loaded automatically, and vise versa. However, if the client
+requests `es` but only `Whatever.es-ES.xlf` is available, `Whatever.es-ES.xlf` will be loaded automatically, and vice versa. However, if the client
 requests `es` and both `Whatever.es-ES.xlf` and `Whatever.es-MX.xlf` are available, or if no `Whatever.es[-details].xlf` is available, a dialog will
 inform the user that the selected language is not available and prompt the user to select from the available languages.
 
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ To use L10NSharp in your application, simply call the `Create` method on `Locali
 passing the location of the translation memory files and some other information:
 
 ```csharp
-using (var lm = LocalizationManager.Create(TranslationMemory.XLiff, lang, "SampleApp",
+using (var lm = LocalizationManager.Create(lang, "SampleApp",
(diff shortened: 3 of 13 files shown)


tombogle (changes requested):
I think you'll want an entry in CHANGELOG.md for this somewhat subtle change.

tombogle:
*[`src/L10NSharp/XLiffUtils/XLiffLocalizedStringCache.cs` line 179 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs3Ejk5Z1rN3OFsahI:-NPs3Ejk5Z1rN3OFsahJ:bw2f3x3) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharp/XLiffUtils/XLiffLocalizedStringCache.cs#L179)):*
> ```Smalltalk
> 				// Another possibility is that the lang folder is "es-ES" but the client is requesting only "es".
> 				// TODO: actual implementation here:
> 				if (langId == "zz") return;
> ```

Although there is one exception, it seems our normal standard is to put the return on a line by itself. IMHO, that makes for more readable code. It's easy to miss the return at the end of the line when scanning through the code quickly.

tombogle:
*[`src/L10NSharp/XLiffUtils/XLiffLocalizedStringCache.cs` line 178 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs4aSfBvmO4jO_l0NL:-NPs4aSg8H7qw6Vs0lLc:bcbqekp) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharp/XLiffUtils/XLiffLocalizedStringCache.cs#L178)):*
> ```Smalltalk
> 					return;
> 				// Another possibility is that the lang folder is "es-ES" but the client is requesting only "es".
> 				// TODO: actual implementation here:
> ```

Can you expand on what this TODO is about? I'm guessing the idea is that someday we might know of specific language tag for which we can do something better than just give up. Is that the idea?

tombogle:
*[`src/L10NSharpTests/LocalizationManagerTestsBase.cs` line 438 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs5mCcA_OI4IUut0tl:-NPs5mCcA_OI4IUut0tm:b-kdd6rm) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/LocalizationManagerTestsBase.cs#L438)):*
> ```Smalltalk
> 		[TestCase(new[] { "en" }, "blahInEnglishCode", "en", TestName = "en; finds en")]
> 		[TestCase(new[] { "en", "fr" }, "blahInEnglishCode", "en", TestName = "en,fr; finds en")]
> 		[TestCase(new[] { "ar", "en" }, "blahInEnglishCode", "en", TestName = "ar,en; finds en")] // our arabic doesn't have a translation of 'blah', so fall to the code's English
> ```

Not your fault, but this should say "fall back"

tombogle:
*[`src/L10NSharpTests/LocalizationManagerTestsBase.cs` line 581 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs6Yb29rQPPf1eyl3k:-NPs6Yb34YrEyBnHX2vT:b-5gqjju) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/LocalizationManagerTestsBase.cs#L581)):*
> ```Smalltalk
> 		}
> 
> 		private void AddChineseChineseTranslation(string folderPath)
> ```

Not sure I understand the purpose of having Chinese twicetwice in the method name. zh-CN is Simplified Chinese as used in China. Perhaps the name should be AddSimplifiedChineseTranslation OR AddSimplifiedChineseOfChinaTranslation OR AddChineseOfChinaTranslation OR AddZhCnChineseTranslation OR AddChineseZhCnTranslation OR AddChineseTranslation.

tombogle:
*[`src/L10NSharpTests/XLiffLocalizedStringCacheTests.cs` line 86 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPsC7Fk85i0f05THMjs:-NPsC7Fl5djE6BH5m2y7:b38udxe) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/XLiffLocalizedStringCacheTests.cs#L86)):*
> ```Smalltalk
> 		public void TryGetDocument()
> 		{
> 			//var sut = new XLiffLocalizedStringCache();
> ```

Is this unfinished work, or something that needs to be reverted?

tombogle (commented):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/110)* status: 3 of 4 files reviewed, 6 unresolved discussions (waiting on @papeh)

papeh (commented):
> Dismissed @tombogle from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/110)* status: 1 of 4 files reviewed, 5 unresolved discussions (waiting on @papeh and @tombogle)

___
*[`src/L10NSharp/XLiffUtils/XLiffLocalizedStringCache.cs` line 178 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs4aSfBvmO4jO_l0NL:-NPsu7kM76ETjD9x8fZL:b3tar7t) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharp/XLiffUtils/XLiffLocalizedStringCache.cs#L178)):*

Can you expand on what this TODO is about? I'm guessing the idea is that someday we might know of specific language tag for which we can do something better than just give up. Is that the idea?
</blockquote></details>

I plan to implement the behaviour for the remaining cases before removing this PR's draft status.
___
*[`src/L10NSharp/XLiffUtils/XLiffLocalizedStringCache.cs` line 179 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs3Ejk5Z1rN3OFsahI:-NPstwVm3lAHApUeY1tG:b-u39x5x) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharp/XLiffUtils/XLiffLocalizedStringCache.cs#L179)):*

Although there is one exception, it seems our normal standard is to put the return on a line by itself. IMHO, that makes for more readable code. It's easy to miss the return at the end of the line when scanning through the code quickly.
</blockquote></details>

That line was temporary. It has been removed.
___
*[`src/L10NSharpTests/LocalizationManagerTestsBase.cs` line 438 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs5mCcA_OI4IUut0tl:-NPsujCA0B8vnJMrcb7Q:b-d7bhn0) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/LocalizationManagerTestsBase.cs#L438)):*

Not your fault, but this should say "fall back"
</blockquote></details>

will do
___
*[`src/L10NSharpTests/LocalizationManagerTestsBase.cs` line 581 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs6Yb29rQPPf1eyl3k:-NPsuga_BqZ-_76jLx9I:bgvs52y) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/LocalizationManagerTestsBase.cs#L581)):*

Not sure I understand the purpose of having Chinese twicetwice in the method name. zh-CN is Simplified Chinese as used in China. Perhaps the name should be AddSimplifiedChineseTranslation OR AddSimplifiedChineseOfChinaTranslation OR AddChineseOfChinaTranslation OR AddZhCnChineseTranslation OR AddChineseZhCnTranslation OR AddChineseTranslation.
</blockquote></details>

will do.
___
*[`src/L10NSharpTests/LocalizationManagerTestsBase.cs` line 987 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs9IcxE6exdm8InWOK:-NPszIgSEKJ9huUm2gwO:bq92tas) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/LocalizationManagerTestsBase.cs#L987)):*

I know you said you intended to leaved "undefined" the behavior when "zn" is requested, but multiple zn- variants are available. Specifically in the case of "zn", I'm not sure this is a good idea. I don't know when/where we expect an application to request "zn", but I think it will be very common that if there is an available "zn-" language, there will be at least two. If it is truly impossible to decide upon a defined order of precedence, then I can't think of a benefit of implementing this behavior. If a caller can't know what to expect, why would they call the method?
</blockquote></details>

How about I add a test that it returns a match only if there is exactly one?
___
*[`src/L10NSharpTests/XLiffLocalizedStringCacheTests.cs` line 86 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPsC7Fk85i0f05THMjs:-NPszeTG2FOPW8JMzA0X:bj1nvfx) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/XLiffLocalizedStringCacheTests.cs#L86)):*

Is this unfinished work, or something that needs to be reverted?
</blockquote></details>

unfinished work.

papeh (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/110)* status: 1 of 4 files reviewed, 5 unresolved discussions (waiting on @tombogle)

___
*[`src/L10NSharpTests/LocalizationManagerTestsBase.cs` line 987 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs9IcxE6exdm8InWOK:-NPt-_tMEyK_Yk-H9ts4:brvwnzm) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/LocalizationManagerTestsBase.cs#L987)):*
<details><summary><i>Previously, papeh wrote…</i></summary><blockquote>

How about I add a test that it returns a match only if there is exactly one?
</blockquote></details>

(FieldWorks ships only zh-CN, because some systems (Mono and historically Windows) do not recognize `zh` on its own.

tombogle:
*[`src/L10NSharpTests/LocalizationManagerTestsBase.cs` line 987 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs9IcxE6exdm8InWOK:-NPwZWyq5t1v6ludaz35:b-9s9gz0) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/LocalizationManagerTestsBase.cs#L987)):*
<details><summary><i>Previously, papeh wrote…</i></summary><blockquote>

(FieldWorks ships only zh-CN, because some systems (Mono and historically Windows) do not recognize `zh` on its own.
</blockquote></details>

I think that seems like a reasonable implementation (at least until we can find something more reasonable to do in some case).

papeh (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/110)* status: 1 of 5 files reviewed, 5 unresolved discussions (waiting on @tombogle)

___
*[`src/L10NSharpTests/LocalizationManagerTestsBase.cs` line 438 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs5mCcA_OI4IUut0tl:-NPt6QIc5_u5IgPREKzm:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/LocalizationManagerTestsBase.cs#L438)):*
<details><summary><i>Previously, papeh wrote…</i></summary><blockquote>

will do
</blockquote></details>

Done.
___
*[`src/L10NSharpTests/LocalizationManagerTestsBase.cs` line 581 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPs6Yb29rQPPf1eyl3k:-NPt6TGE-YV4bQ1NJ4UI:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/LocalizationManagerTestsBase.cs#L581)):*
<details><summary><i>Previously, papeh wrote…</i></summary><blockquote>

will do.
</blockquote></details>

Done.

papeh (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/110)* status: 0 of 14 files reviewed, 5 unresolved discussions (waiting on @tombogle)

___
*[`src/L10NSharpTests/XLiffLocalizedStringCacheTests.cs` line 86 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPsC7Fk85i0f05THMjs:-NPyHlLHD5D8fFPwG6pI:bjiwfnb) ([raw file](https://github.com/sillsdev/l10nsharp/blob/5aa9dbd4c71d7d317749b28a7fedcfa889ecff07/src/L10NSharpTests/XLiffLocalizedStringCacheTests.cs#L86)):*
<details><summary><i>Previously, papeh wrote…</i></summary><blockquote>

unfinished work.
</blockquote></details>

this part of XLiffLocalizedStringCache is easier to test in LocalizationManagerTestsBase

tombogle:
*[`CHANGELOG.md` line 23 at r6](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPzK6qh5rU9J9h3JQAI:-NPzK6qh5rU9J9h3JQAJ:b-yjn4f3) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a833f7dce1e4ee707b4ecd40fdf9222fc368279b/CHANGELOG.md#L23)):*
> ```Markdown
> - `LocalizationManager.Create` methods without `TranslationMemory kind` parameter
> 
> ### Fixed
> ```

Even if no public members were changed, I wonder if we want this to be regarded as a "fix" or a "change"? Not 100% sure that these changes wouldn't be in some sense "breaking", given the new behavior. Hopefully not, but it's hard to be sure.

tombogle:
*[`L10NSharp.sln.DotSettings` line 7 at r6](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPzM7ku-3tXhPu-NMbN:-NPzM7ku-3tXhPu-NMbO:b-f3ocv5) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a833f7dce1e4ee707b4ecd40fdf9222fc368279b/L10NSharp.sln.DotSettings#L7)):*
> ```
> 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Filenames/@EntryIndexedValue">True</s:Boolean>
> 	<s:Boolean x:Key="/Default/UserDictionary/Words/=langs/@EntryIndexedValue">True</s:Boolean>
> 	<s:Boolean x:Key="/Default/UserDictionary/Words/=Liff/@EntryIndexedValue">True</s:Boolean>
> ```

I had considered doing this at one point, but since Liff isn't really a word and the miscapitalization is something that probably "should" be fixed if it weren't so entrenched, I kind of wondered if it was maybe better to just let it keep complaining.

tombogle (commented):
Just a few outstanding questions to ponder. Otherwise, LGTM. Since a lot of projects depend on this, might not be all bad to get another set of eyes to review it, maybe someone from the Bloom team. 

> Reviewed 2 of 4 files at r4, 3 of 12 files at r5, 11 of 11 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-:-NPzPG0sCQqw4Ic_Dk_I:b-y1t6rt)* status: all files reviewed, 2 unresolved discussions (waiting on @papeh)

papeh (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/110)* status: 5 of 14 files reviewed, 2 unresolved discussions (waiting on @tombogle)

___
*[`CHANGELOG.md` line 23 at r6](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPzK6qh5rU9J9h3JQAI:-NQ0t1yzESIJwsMZKHSU:b-i0lt1c) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a833f7dce1e4ee707b4ecd40fdf9222fc368279b/CHANGELOG.md#L23)):*

Even if no public members were changed, I wonder if we want this to be regarded as a "fix" or a "change"? Not 100% sure that these changes wouldn't be in some sense "breaking", given the new behavior. Hopefully not, but it's hard to be sure.
</blockquote></details>

Yes, it breaks the behaviour where creating a manager for "es" when only "es-ES" is available results in a user prompt. The only documentation I found (comment on LocalizationManagerInternal<T>.MapToExistingLanguageIfPossible) says that the post-PR behaviour is correct (at least after es-ES has been loaded).
___
*[`L10NSharp.sln.DotSettings` line 7 at r6](https://reviewable.io/reviews/sillsdev/l10nsharp/110#-NPzM7ku-3tXhPu-NMbN:-NQ19HexDixhhdjum2IL:b-ldwsbj) ([raw file](https://github.com/sillsdev/l10nsharp/blob/a833f7dce1e4ee707b4ecd40fdf9222fc368279b/L10NSharp.sln.DotSettings#L7)):*

I had considered doing this at one point, but since Liff isn't really a word and the miscapitalization is something that probably "should" be fixed if it weren't so entrenched, I kind of wondered if it was maybe better to just let it keep complaining.
</blockquote></details>

I was about to fix them, but so many are public, and I couldn't be bothered to deprecate them and set up aliases.

Issue #116, opened 2024-02-21
L10NSharp is a .Net framework only package but is depended on by .Net Standard packages 


hahn-kev:
For example [SIL.ChorusPlugin.LfMergeBridge](https://www.nuget.org/packages/SIL.ChorusPlugin.LfMergeBridge) depnds on L10NSharp despite it being a .Net Standard package. This causes warnings like this every time you run a package restore:
```
warning NU1701: Package 'L10NSharp 6.0.0' was restored using
'.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8, .NETFramework,Version=v4.8.1'
instead of the project target framework 'net8.0'. This package may not be fully compatible with your project.
```

taking a look at the NuGet dependancies there's no reason we couldn't upgrade this to support .Net standard is that correct?

ermshiperete:
It uses System.Windows.Forms which I think isn't available in .NET standard.

tombogle:
It should be possible to refactor this (as we did years ago with the libpalaso DLLs) so that the WinForms stuff is in a separate DLL with appropriate interfaces and logic to hook up UI stuff on Windows. I don't think the core localization logic is dependent on WinForms.

hahn-kev:
> It uses System.Windows.Forms which I think isn't available in .NET standard.

Forms are available in dotnet 5+, so it could multitarget framework and a modern version of dotnet with Win Forms.

hatton:
@tombogle do you know if anything is using the UI parts? One option would be to just drop those and ask any clients using them to roll them in locally.

tombogle:
Yes, the UI stuff is definitely being used by a large number of applications. You might be able to make a (weak?) case for requiring each of the applications to have their own implementations of the dialog boxes, but the L10nExtender is a pretty fundamental piece of the localization puzzle for the WinForms clients, so I think it would be a step backward to require each of them to implement that and maintain separate versions.

megahirt:
I'm thinking that what is needed is for us to understand why SIL.ChorusPlugin.LfMergeBridge needs L10NSharp in the first place and refactor that dependency out.

imnasnainaec:
[...]

PR #118, opened 2024-04-12
Adding Framework 4.8 and .NET8 to targets.

josephmyers:
This required platform-specific references in the project files. I also fixed a warning about Assembly.CodeBase and some .NET8 warnings about manually copying duplicate files into our nuget package folder when packing. These files appear to be duplicates of what is copied during the .NET8 packaging.

The only other change worth noting involves the "unknown language" detection in L10nCultureInfo. It appears, per [the link listed in the comment](https://stackoverflow.com/a/71388328/1964319) at that point in the code, that our method of determining an unknown language no longer works, starting with Win10 (combined with .NET). The change I made should be compatible with older OS's, like Win7, but I don't have an easy way to confirm this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)

### DIFF
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -22,7 +22,7 @@ jobs:
     name: "Build and Test"
     strategy:
       matrix:
-        os: [windows-latest]
+        os: [windows-latest, ubuntu-latest]
     runs-on: ${{ matrix.os }}
 
     steps:
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -20,7 +20,10 @@ jobs:
 
   build-and-test:
     name: "Build and Test"
-    runs-on: windows-latest
+    strategy:
+      matrix:
+        os: [windows-latest]
+    runs-on: ${{ matrix.os }}
 
     steps:
     - name: Checkout
@@ -36,8 +39,13 @@ jobs:
     - name: Build and Pack
       run: dotnet build --configuration Release
 
+    - name: Test (Ubuntu)
+      run: dotnet test --configuration Release --no-build --filter TestCategory!=RequiresDisplay -- NUnit.TestOutputXml=TestResults
+      if: matrix.os == 'ubuntu-latest'
+
     - name: Test (Windows)
       run: dotnet test --configuration Release --no-build -- NUnit.TestOutputXml=TestResults
+      if: matrix.os != 'ubuntu-latest'
 
     - name: Upload Test Results
       if: always()
@@ -53,6 +61,7 @@ jobs:
         path: |
           output/*.nupkg
           output/*.snupkg
+      if: matrix.os == 'ubuntu-latest'
 
   publish-nuget:
     name: "Publish NuGet package"
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -39,10 +39,6 @@ jobs:
     - name: Build and Pack
       run: dotnet build --configuration Release
 
-    - name: Test (Ubuntu)
-      run: dotnet test --configuration Release --no-build output/Release/net461/*Tests.dll --filter TestCategory!=RequiresDisplay -- NUnit.TestOutputXml=TestResults
-      if: matrix.os == 'ubuntu-latest'
-
     - name: Test (Windows)
       run: dotnet test --configuration Release --no-build -- NUnit.TestOutputXml=TestResults
       if: matrix.os != 'ubuntu-latest'
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -40,7 +40,7 @@ jobs:
       run: dotnet build --configuration Release
 
     - name: Test (Ubuntu)
-      run: dotnet test --configuration Release --no-build output\Release\net461\*Tests.dll --filter TestCategory!=RequiresDisplay -- NUnit.TestOutputXml=TestResults
+      run: dotnet test --configuration Release --no-build output/Release/net461/*Tests.dll --filter TestCategory!=RequiresDisplay -- NUnit.TestOutputXml=TestResults
       if: matrix.os == 'ubuntu-latest'
 
     - name: Test (Windows)
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -44,7 +44,6 @@ jobs:
       if: matrix.os == 'ubuntu-latest'
 
     - name: Test (Windows)
-      # specify path to test assembly as workaround for https://github.com/nunit/nunit3-vs-adapter/issues/1040
       run: dotnet test --configuration Release --no-build -- NUnit.TestOutputXml=TestResults
       if: matrix.os != 'ubuntu-latest'
 
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -40,7 +40,7 @@ jobs:
       run: dotnet build --configuration Release
 
     - name: Test (Ubuntu)
-      run: dotnet test --configuration Release --no-build --filter TestCategory!=RequiresDisplay -- NUnit.TestOutputXml=TestResults
+      run: dotnet test --configuration Release --no-build output\Release\net461\*Tests.dll --filter TestCategory!=RequiresDisplay -- NUnit.TestOutputXml=TestResults
       if: matrix.os == 'ubuntu-latest'
 
     - name: Test (Windows)
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -34,7 +34,7 @@ jobs:
     - name: Install .NET Core
       uses: actions/setup-dotnet@c0d4ad69d8bd405d234f1c9166d383b7a4f69ed8 # v2.1.0
       with:
-        dotnet-version: 7.0.x
+        dotnet-version: 8.0.x
 
     - name: Build and Pack
       run: dotnet build --configuration Release
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -22,7 +22,7 @@ jobs:
     name: "Build and Test"
     strategy:
       matrix:
-        os: [windows-latest, ubuntu-latest]
+        os: [windows-latest]
     runs-on: ${{ matrix.os }}
 
     steps:
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -39,7 +39,11 @@ jobs:
     - name: Build and Pack
       run: dotnet build --configuration Release
 
-    - name: Test (Windows)
+    - name: Test (Framework 4.6.1)
+      run: dotnet test --configuration Release --no-build output\Release\net461\*Tests.dll -- NUnit.TestOutputXml=TestResults
+      if: matrix.os != 'ubuntu-latest'
+
+    - name: Test (Framework and .NET latest)
       run: dotnet test --configuration Release --no-build -- NUnit.TestOutputXml=TestResults
       if: matrix.os != 'ubuntu-latest'
 
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -45,7 +45,7 @@ jobs:
 
     - name: Test (Windows)
       # specify path to test assembly as workaround for https://github.com/nunit/nunit3-vs-adapter/issues/1040
-      run: dotnet test --configuration Release --no-build output\Release\net461\*Tests.dll -- NUnit.TestOutputXml=TestResults
+      run: dotnet test --configuration Release --no-build -- NUnit.TestOutputXml=TestResults
       if: matrix.os != 'ubuntu-latest'
 
     - name: Upload Test Results
--- a/.github/workflows/CI-CD.yml
+++ b/.github/workflows/CI-CD.yml
@@ -20,10 +20,7 @@ jobs:
 
   build-and-test:
     name: "Build and Test"
-    strategy:
-      matrix:
-        os: [windows-latest]
-    runs-on: ${{ matrix.os }}
+    runs-on: windows-latest
 
     steps:
     - name: Checkout
@@ -39,13 +36,8 @@ jobs:
     - name: Build and Pack
       run: dotnet build --configuration Release
 
-    - name: Test (Ubuntu)
-      run: dotnet test --configuration Release --no-build --filter TestCategory!=RequiresDisplay -- NUnit.TestOutputXml=TestResults
-      if: matrix.os == 'ubuntu-latest'
-
     - name: Test (Windows)
       run: dotnet test --configuration Release --no-build -- NUnit.TestOutputXml=TestResults
-      if: matrix.os != 'ubuntu-latest'
 
     - name: Upload Test Results
       if: always()
@@ -61,7 +53,6 @@ jobs:
         path: |
           output/*.nupkg
           output/*.snupkg
-      if: matrix.os == 'ubuntu-latest'
 
   publish-nuget:
     name: "Publish NuGet package"
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <TargetFrameworks>net461;net481;net8.0-windows</TargetFrameworks>
+    <TargetFrameworks>net461;net48;net8.0-windows</TargetFrameworks>
     <Company>SIL</Company>
     <Authors>SIL International</Authors>
     <Product>L10NSharp</Product>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -25,6 +25,7 @@ See full changelog at https://github.com/sillsdev/l10nsharp/blob/master/CHANGELO
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
     <UseWindowsForms>true</UseWindowsForms>
+    <EnableWindowsTargeting>true</EnableWindowsTargeting>
   </PropertyGroup>
   <ItemGroup>
     <!-- Without this line some projects fail to build on TC with "error : SourceRoot items
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,6 @@
 <Project>
   <PropertyGroup>
-    <TargetFrameworks>net461</TargetFrameworks>
+    <TargetFrameworks>net461;net481;net8.0-windows</TargetFrameworks>
     <Company>SIL</Company>
     <Authors>SIL International</Authors>
     <Product>L10NSharp</Product>
@@ -24,6 +24,7 @@ See full changelog at https://github.com/sillsdev/l10nsharp/blob/master/CHANGELO
     </AppendToReleaseNotesProperty>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
+    <UseWindowsForms>true</UseWindowsForms>
   </PropertyGroup>
   <ItemGroup>
     <!-- Without this line some projects fail to build on TC with "error : SourceRoot items
--- a/src/CheckOrFixXliff/CheckOrFixXliff.csproj
+++ b/src/CheckOrFixXliff/CheckOrFixXliff.csproj
@@ -9,9 +9,9 @@
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/sillsdev/l10nsharp/tree/master/src/CheckOrFixXliff</PackageProjectUrl>
     <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);CopyProjectReferencesToPackage</TargetsForTfmSpecificBuildOutput>
-    <NoWarn>$(NoWarn);NU5128</NoWarn>
+    <NoWarn>$(NoWarn);NU5128;NU5118</NoWarn>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' Or '$(TargetFramework)' == 'net481' ">
     <Reference Include="System.ServiceModel" />
     <Reference Include="System.Windows.Forms" />
   </ItemGroup>
--- a/src/CheckOrFixXliff/CheckOrFixXliff.csproj
+++ b/src/CheckOrFixXliff/CheckOrFixXliff.csproj
@@ -11,7 +11,7 @@
     <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);CopyProjectReferencesToPackage</TargetsForTfmSpecificBuildOutput>
     <NoWarn>$(NoWarn);NU5128;NU5118</NoWarn>
   </PropertyGroup>
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' Or '$(TargetFramework)' == 'net481' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' Or '$(TargetFramework)' == 'net48' ">
     <Reference Include="System.ServiceModel" />
     <Reference Include="System.Windows.Forms" />
   </ItemGroup>
--- a/src/CheckOrFixXliff/Program.cs
+++ b/src/CheckOrFixXliff/Program.cs
@@ -176,7 +176,7 @@ namespace CheckOrFixXliff
 		private static bool ValidateXliffAgainstSchema(string filename)
 		{
 			bool valid = true;
-			var installedXliffDir = Path.GetDirectoryName(new Uri(Assembly.GetExecutingAssembly().CodeBase).LocalPath);
+			var installedXliffDir = Path.GetDirectoryName(new Uri(Assembly.GetExecutingAssembly().Location).LocalPath);
 			var schemaLocation = Path.Combine(installedXliffDir, "xliff-core-1.2-transitional.xsd");
 			var schemas = new XmlSchemaSet();
 			using (var reader = XmlReader.Create(schemaLocation))
--- a/src/ExtractXliff/ExtractXliff.csproj
+++ b/src/ExtractXliff/ExtractXliff.csproj
@@ -11,7 +11,7 @@
     <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);CopyProjectReferencesToPackage</TargetsForTfmSpecificBuildOutput>
     <NoWarn>$(NoWarn);NU5128;NU5118</NoWarn>
   </PropertyGroup>
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' Or '$(TargetFramework)' == 'net481' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' Or '$(TargetFramework)' == '' ">
     <Reference Include="System.ServiceModel" />
     <Reference Include="System.Windows.Forms" />
   </ItemGroup>
--- a/src/ExtractXliff/ExtractXliff.csproj
+++ b/src/ExtractXliff/ExtractXliff.csproj
@@ -9,9 +9,9 @@
     <PackageId>L10NSharp.ExtractXliff</PackageId>
     <PackageProjectUrl>https://github.com/sillsdev/l10nsharp/tree/master/src/ExtractXliff</PackageProjectUrl>
     <TargetsForTfmSpecificBuildOutput>$(TargetsForTfmSpecificBuildOutput);CopyProjectReferencesToPackage</TargetsForTfmSpecificBuildOutput>
-    <NoWarn>$(NoWarn);NU5128</NoWarn>
+    <NoWarn>$(NoWarn);NU5128;NU5118</NoWarn>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net461' Or '$(TargetFramework)' == 'net481' ">
     <Reference Include="System.ServiceModel" />
     <Reference Include="System.Windows.Forms" />
   </ItemGroup>
--- a/src/L10NSharp/L10NCultureInfo.cs
+++ b/src/L10NSharp/L10NCultureInfo.cs
@@ -34,7 +34,13 @@ namespace L10NSharp
 			{
 				RawCultureInfo = null;
 			}
-			if (RawCultureInfo == null || RawCultureInfo.EnglishName.StartsWith("Unknown Language"))
+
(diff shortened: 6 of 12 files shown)


megahirt (commented):
so far so good, Joseph!  It looks like the strategy we're using at this point is to have conditional dependencies based upon the target build.

> Reviewed 7 of 7 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-:-NvFgw_6DewcSDQEGzix:bpdj3xd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers:
I need to lodge a version bump with semvar

megahirt (approved):
LGTM!  Going forward, I am a proponent of not building L10NSharp on Ubuntu Mono since the purpose is for use in .Net Framework applications which are Windows only.

> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-:-NxWzOHTAjaJfPINCkoN:b-sqlz5a)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @andrew-polk)

andrew-polk (changes requested):
> Going forward, I am a proponent of not building L10NSharp on Ubuntu Mono since the purpose is for use in .Net Framework applications which are Windows only.

I'm certainly no expert, but doesn't mono work with .NET Framework? 
@ermshiperete 

> Reviewed 6 of 7 files at r1, 5 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-:-NxY7cZ-DR6gWqygWdlX:b-32we6q)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`src/L10NSharp/L10NCultureInfo.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8L8y35Yjt22poWQ3:-NxY8L8y35Yjt22poWQ4:b-n1ncyw) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NCultureInfo.cs#L39)):*
> ```Smalltalk
> 
> 			// Starting with Windows 10, unknown cultures no longer return a RawCultureInfo containing an "Unknown Language" indication.
> 			// The proper way to detect fully unknown cultures, for Windows 10 and prior, is to:
> ```

Something in this comment is confusing me.
Is it supposed to be "for Windows 10 and later"?
___
*[`src/L10NSharp/L10NCultureInfo.cs` line 43 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8bMQD7_JrsI4ag7A:-NxY8bMQD7_JrsI4ag7B:b-fbj8eg) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NCultureInfo.cs#L43)):*
> ```Smalltalk
> 			//    2. Check if the three-letter language name is set to default
> 			var isFullyUnknown = RawCultureInfo.CultureTypes.HasFlag(CultureTypes.UserCustomCulture) && RawCultureInfo.ThreeLetterWindowsLanguageName == "ZZZ";
> 			if (RawCultureInfo == null || isFullyUnknown)
> ```

To maintain backward compatibility, should we still include the "Unknown Language" check?
I guess maybe we don't care...
___
*[`src/L10NSharp/L10NSharp.csproj` line 13 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8qsnECdy3GrufM91:-NxY8qsnECdy3GrufM92:b-lgr1ym) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NSharp.csproj#L13)):*
> ```XML
>     <Reference Include="System.Windows.Forms" />
>   </ItemGroup>
>   <ItemGroup Condition=" '$(TargetFramework)' == 'net461' Or '$(TargetFramework)' == 'net481' ">
> ```

Maybe this was the result of an automated change, but the result is that the four items in this group are duplicated in two different conditional groups. (When they could just be unconditional.)
___
*[`src/L10NSharpTests/L10NCultureInfoTests.cs` line 45 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY9aKtEjZu5pecMP5M:-NxY9aKtEjZu5pecMP5N:bnv78j7) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharpTests/L10NCultureInfoTests.cs#L45)):*
> ```Smalltalk
> 				Assert.AreEqual("pbu", pbuci.RawCultureInfo.Name);
> 				Assert.IsTrue(pbuci.RawCultureInfo.CultureTypes.HasFlag(CultureTypes.UserCustomCulture));
> 				Assert.AreEqual("ZZZ", pbuci.RawCultureInfo.ThreeLetterWindowsLanguageName);
> ```

What is `pbuci.RawCultureInfo.EnglishName` now? Still something reasonable?

josephmyers (commented):
Yes, Mono will only work with .NET Framework, and we're moving away from that, as discussed in the .NET8 migration Slack channel. Granted, we're not there yet, hence the three frameworks!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-:-Ny3QLl08WylbbCmyjj7:b-locqko)* status: 11 of 12 files reviewed, 4 unresolved discussions (waiting on @andrew-polk and @megahirt)

___
*[`src/L10NSharp/L10NCultureInfo.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8L8y35Yjt22poWQ3:-Ny3QwjF3ztFWtREa9JJ:b77lanw) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NCultureInfo.cs#L39)):*

Something in this comment is confusing me.
Is it supposed to be "for Windows 10 and later"?
</blockquote></details>

Windows 10 and prior is correct. I wanted to be backwards compatible, so I did some research (in lieu of testing on an old OS), with this as my primary finding:
https://stackoverflow.com/a/71388328/1964319
___
*[`src/L10NSharp/L10NCultureInfo.cs` line 43 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8bMQD7_JrsI4ag7A:-Ny3SVVZ579IdItRdBBy:b6wcyrd) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NCultureInfo.cs#L43)):*

To maintain backward compatibility, should we still include the "Unknown Language" check?
I guess maybe we don't care...
</blockquote></details>

Ah, see above comment. The new check theoretically covers both cases.
___
*[`src/L10NSharp/L10NSharp.csproj` line 13 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8qsnECdy3GrufM91:-Ny3T9R10enbx2Nzy6yE:bonsh7j) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NSharp.csproj#L13)):*

Maybe this was the result of an automated change, but the result is that the four items in this group are duplicated in two different conditional groups. (When they could just be unconditional.)
</blockquote></details>

Thanks, that's a good catch!
___
*[`src/L10NSharpTests/L10NCultureInfoTests.cs` line 45 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY9aKtEjZu5pecMP5M:-Ny3UANh4iValO62mJ3t:b-b7mtuo) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharpTests/L10NCultureInfoTests.cs#L45)):*

What is `pbuci.RawCultureInfo.EnglishName` now? Still something reasonable?
</blockquote></details>

It's not consistent between frameworks. For Framework, it's the original "Unknown Language (pbu)" value, and for .NET8, it's just "pbu." The check I'm using works the same between the two.

andrew-polk (changes requested):
> Reviewed 1 of 6 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`.github/workflows/CI-CD.yml` line 25 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69N5o41f-2zDgHLW5:-Ny69N5o41f-2zDgHLW6:b-qmie2f) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L25)):*
> ```YAML
>     strategy:
>       matrix:
>         os: [windows-latest, ubuntu-latest]
> ```

Sorry; I'm not clear why we are dropping ubuntu at this time.
And I wouldn't be comfortable doing so without @ermshiperete or someone else in that realm approving it.
___
*[`.github/workflows/CI-CD.yml` line 48 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69G5YBt6B8VJPyeT_:-Ny69G5YBt6B8VJPyeTa:b-2rpdlo) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L48)):*
> ```YAML
>     - name: Test (Windows)
>       # specify path to test assembly as workaround for https://github.com/nunit/nunit3-vs-adapter/issues/1040
>       run: dotnet test --configuration Release --no-build output\Release\net461\*Tests.dll -- NUnit.TestOutputXml=TestResults
> ```

For the 4.6.1 tests, I think with my change to master, we should be able to have those run again.
Does that mean you have to put back the `output\Release\net461\*Tests.dll`?

andrew-polk (changes requested):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`.github/workflows/CI-CD.yml` line 48 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69G5YBt6B8VJPyeT_:-Ny6Ci7NBt7OaqEeuu_o:balvijo) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L48)):*

For the 4.6.1 tests, I think with my change to master, we should be able to have those run again.
Does that mean you have to put back the `output\Release\net461\*Tests.dll`?
</blockquote></details>

But that would have to be conditional, obviously...

andrew-polk (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: all files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/L10NSharp/L10NCultureInfo.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8L8y35Yjt22poWQ3:-Ny6D8ab6YGn2tY51EPb:b-i6xld4) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NCultureInfo.cs#L39)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

Windows 10 and prior is correct. I wanted to be backwards compatible, so I did some research (in lieu of testing on an old OS), with this as my primary finding:
https://stackoverflow.com/a/71388328/1964319
</blockquote></details>

Ok, I'll try to talk out my confusion, and you can help me see what I'm missing.

First, your comment states that "Starting with Win 10..." but then continues with what seems to be a "therefore, we're changing it to this other thing to handle Win10 and earlier." 

Stating what I think is the same confusion but from the perspective of the code change,
it seems that we previously used "Unknown Language" for this check, and you've corrected it for older OSes. But now I'm not understanding how we are handling it for newer OSes.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: 11 of 12 files reviewed, 3 unresolved discussions (waiting on @andrew-polk, @ermshiperete, and @megahirt)

___
*[`.github/workflows/CI-CD.yml` line 25 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69N5o41f-2zDgHLW5:-NyJHlq45bf1YtInixV7:b-7ahpuj) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L25)):*

Sorry; I'm not clear why we are dropping ubuntu at this time.
And I wouldn't be comfortable doing so without @ermshiperete or someone else in that realm approving it.
</blockquote></details>

@megahirt did approve it. Maybe I should be super clear here with him and confirm he's aware of this particular change?

Keep in mind that eberhard has mentioned previously (in Slack) that he's intentionally taking less of a role over these repo's, so I don't think we should rely on decisions from him. Of course his input is valued. Right now Chris is the authoritative word on this code, since he's taking the responsibility of maintaining it.
___
*[`.github/workflows/CI-CD.yml` line 48 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69G5YBt6B8VJPyeT_:-NyJIMJP7lIEKcYQ_1lA:barhsca) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L48)):*

But that would have to be conditional, obviously...
</blockquote></details>

Yes, thanks for that fix! I tried to get them all running again, but even with the downgrade it is proving nontrivial. Removing this allows the 4.8.1 and .NET8 tests to run, which @megahirt deemed acceptable. He confirmed again today.

I also messed around with the wildcard some, but without any luck. If you know of another path, I'm absolutely open to it. But it doesn't appear to be very highly valuable, given Framework 4.8.1 is working.
___
*[`src/L10NSharp/L10NCultureInfo.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8L8y35Yjt22poWQ3:-NyJJ0sLDYyL9zIQxJq2:bxw3awc) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NCultureInfo.cs#L39)):*

Ok, I'll try to talk out my confusion, and you can help me see what I'm missing.

First, your comment states that "Starting with Win 10..." but then continues with what seems to be a "therefore, we're changing it to this other thing to handle Win10 and earlier." 

Stating what I think is the same confusion but from the perspective of the code change,
it seems that we previously used "Unknown Language" for this check, and you've corrected it for older OSes. But now I'm not understanding how we are handling it for newer OSes.
</blockquote></details>

Sure, sorry that my comments are not clear enough.

So, the previous code was working for OS's older than Windows 10. Apparently, Windows 10 changed its behavior for CultureInfo, so we need to note that here. However, research showed that my fix is backwards compatible with older OS's and equivalent (superior) to the previous code, so I wanted to capture that as a comment here, as well.

Of course, that code could change (again) in future Windows versions, so I wouldn't want to put some sort of assumption/guarantee here that it won't. And if it does, I want whomever comes here to understand why I made this change.

Maybe the primary source of confusion arises from my use of "Starting with...," as if I know what's going to happen in the future? I updated the comment to hopefully be clearer.

The climactic argument this change makes is: "The previous way of doing things worked for Win7 and older, but the new way works for Win10 and older." Hopefully this all helps!

andrew-polk (commented):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: all files reviewed, 3 unresolved discussions (waiting on @ermshiperete, @josephmyers, and @megahirt)

___
*[`.github/workflows/CI-CD.yml` line 25 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69N5o41f-2zDgHLW5:-NyMGxnt6xnwrEg64g0q:bvf0kyc) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L25)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

@megahirt did approve it. Maybe I should be super clear here with him and confirm he's aware of this particular change?

Keep in mind that eberhard has mentioned previously (in Slack) that he's intentionally taking less of a role over these repo's, so I don't think we should rely on decisions from him. Of course his input is valued. Right now Chris is the authoritative word on this code, since he's taking the responsibility of maintaining it.
</blockquote></details>

See comment about 4.6.1 tests.
___
*[`.github/workflows/CI-CD.yml` line 48 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69G5YBt6B8VJPyeT_:-NyMG0Z59HaBFXRESyr8:b-cmj9jn) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L48)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

Yes, thanks for that fix! I tried to get them all running again, but even with the downgrade it is proving nontrivial. Removing this allows the 4.8.1 and .NET8 tests to run, which @megahirt deemed acceptable. He confirmed again today.

I also messed around with the wildcard some, but without any luck. If you know of another path, I'm absolutely open to it. But it doesn't appear to be very highly valuable, given Framework 4.8.1 is working.
</blockquote></details>

Sorry; I definitely don't want to be the bottleneck to you moving forward on this.
If @megahirt has thought through the implications and wants to take responsibility for any fallout, I don't think it is my place to stand in the way.
But I do think we have a philosophical difference. In my mind, we should either say we support 4.6.1 and mono or we don't. If we say we support it, I don't think we should be removing anything (within reason) which "proves" we support it.

Feel free to merge.
___
*[`src/L10NSharp/L10NCultureInfo.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8L8y35Yjt22poWQ3:-NyMHUTQ5r6GUE0bfUd9:beumfhs) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NCultureInfo.cs#L39)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

Sure, sorry that my comments are not clear enough.

So, the previous code was working for OS's older than Windows 10. Apparently, Windows 10 changed its behavior for CultureInfo, so we need to note that here. However, research showed that my fix is backwards compatible with older OS's and equivalent (superior) to the previous code, so I wanted to capture that as a comment here, as well.

Of course, that code could change (again) in future Windows versions, so I wouldn't want to put some sort of assumption/guarantee here that it won't. And if it does, I want whomever comes here to understand why I made this change.

Maybe the primary source of confusion arises from my use of "Starting with...," as if I know what's going to happen in the future? I updated the comment to hopefully be clearer.

The climactic argument this change makes is: "The previous way of doing things worked for Win7 and older, but the new way works for Win10 and older." Hopefully this all helps!
</blockquote></details>

I'm still not sure I get it, but I'm going to get out of the way.

If it is a helpful question, I think what I still don't get is, What does Win11 do? 
Or are you considering Win11 to be part of Win10? If yes, that would definitely clear up the confusion.
Anyway, feel free to answer in the PR or a comment. Or neither.

megahirt (approved):
> Reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: all files reviewed, 3 unresolved discussions (waiting on @ermshiperete and @josephmyers)

___
*[`.github/workflows/CI-CD.yml` line 48 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69G5YBt6B8VJPyeT_:-NyOD1VY1VqrLjw39j_i:bwjqpbr) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L48)):*

Sorry; I definitely don't want to be the bottleneck to you moving forward on this.
If @megahirt has thought through the implications and wants to take responsibility for any fallout, I don't think it is my place to stand in the way.
But I do think we have a philosophical difference. In my mind, we should either say we support 4.6.1 and mono or we don't. If we say we support it, I don't think we should be removing anything (within reason) which "proves" we support it.

Feel free to merge.
</blockquote></details>

You've got a good point, @andrew-polk .  When I surveyed all developers on #common-libraries, no one could say that we are still actively supporting/shipping software on linux/mono.  Thus, it seems safe to me to remove complexity in our shared libraries like this one that has been added for linux/mono support.  Reducing complexity will help a lot for any future effort to move toward .Net 8 Windows and .Net cross-platform.

@ermshiperete should have a voice here.  Can you comment?

I can see how Framework 4.6.1 and Mono are tied together for you - we've done all the work, why shouldn't we continue to keep it alive?  I guess for the reason above, I don't see a compelling reason to keep the extra complexity around.

As long as we have a branch that represents the 4.6.1 / mono shipping version, which can stick around forever IMO, we should feel free to move forward with just 4.8.1 on Windows only.  This PR represents forward work toward the 4.8.1 future that is Windows only.

I will also look into making these changes as part of a new long-term branch for .net8

rmunn:
I just checked the Microsoft support site, and Framework versions 4.7.1, 4.7.2, and even 4.8 all have offline installers available for Windows versions as old as Windows 7 SP1. E.g., https://support.microsoft.com/en-us/topic/microsoft-net-framework-4-8-offline-installer-for-windows-9d23f658-3b97-68ab-d013-aa3c3e7495e0 lists supported Windows versions as Windows 7 SP1, Windows 8.1, Windows 10 version 1607 or later, plus a lot of Windows Server versions. [Microsoft's announcement of 4.8.1](https://devblogs.microsoft.com/dotnet/announcing-dotnet-framework-481/) claimed that it was supported on Windows 10 and up, not mentioning Windows 7. I have found references elsewhere to Framework 4.8.1 being available for Windows 7, but I haven't been able to prove it. (I may still have a Win7 VM image I can dig up and test that; I'll report my results later).

Therefore, I'd suggest that it might be worth adjusting this PR to target .Net Framework 4.8 rather than 4.8.1. This would still allow libraries to move to .Net Standard 2.0 and be easy to consume from both .Net Framework and .Net 8.0 apps, while still allowing us to produce releases for users who, for one reason or another, are still stuck on older versions of Windows: our software that needs .Net Framework can offer a version with the 4.8 offline installer packaged with it if necessary.

josephmyers (commented):
That's an easy change. Thanks for your input

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-:-NyccKhDAmWO1VBg0vkn:beeih1y)* status: 5 of 12 files reviewed, 3 unresolved discussions (waiting on @andrew-polk, @ermshiperete, and @megahirt)

___
*[`.github/workflows/CI-CD.yml` line 25 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69N5o41f-2zDgHLW5:-NyccbOE-h6qPBB5FOJC:b-zcjvte) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L25)):*

See comment about 4.6.1 tests.
</blockquote></details>

Thanks
___
*[`.github/workflows/CI-CD.yml` line 48 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69G5YBt6B8VJPyeT_:-NycYEQu84d0RTm_jqtq:bx2shfu) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L48)):*
<details><summary><i>Previously, megahirt (Christopher Hirt) wrote…</i></summary><blockquote>

You've got a good point, @andrew-polk .  When I surveyed all developers on #common-libraries, no one could say that we are still actively supporting/shipping software on linux/mono.  Thus, it seems safe to me to remove complexity in our shared libraries like this one that has been added for linux/mono support.  Reducing complexity will help a lot for any future effort to move toward .Net 8 Windows and .Net cross-platform.

@ermshiperete should have a voice here.  Can you comment?

I can see how Framework 4.6.1 and Mono are tied together for you - we've done all the work, why shouldn't we continue to keep it alive?  I guess for the reason above, I don't see a compelling reason to keep the extra complexity around.

As long as we have a branch that represents the 4.6.1 / mono shipping version, which can stick around forever IMO, we should feel free to move forward with just 4.8.1 on Windows only.  This PR represents forward work toward the 4.8.1 future that is Windows only.

I will also look into making these changes as part of a new long-term branch for .net8
</blockquote></details>

We have been discussing a branching strategy for libpalaso [here](https://github.com/sillsdev/libpalaso/issues/1320). I was not planning on applying that to L10nSharp, since other than the 4.6.1 tests and Ubuntu, everything is living happily together. Are you proposing we implement the branching strategy for L10nSharp, as well? That will fundamentally change this PR, and future work on this repo specifically, but it would allow us to more explicitly support the theoretical 4.6.1/Mono userbase.

Maybe these changes would be more appealing if I were actually able to get the ubuntu build to work? This would be regardless of whether we want 4.6.1 on the master branch or an eventual framework branch. I will investigate some more.

(I'm not confident about getting all three sets of tests to run.)
___
*[`src/L10NSharp/L10NCultureInfo.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8L8y35Yjt22poWQ3:-NycaI2w-1YMxuMzUzmA:b-vvjkpk) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NCultureInfo.cs#L39)):*

I'm still not sure I get it, but I'm going to get out of the way.

If it is a helpful question, I think what I still don't get is, What does Win11 do? 
Or are you considering Win11 to be part of Win10? If yes, that would definitely clear up the confusion.
Anyway, feel free to answer in the PR or a comment. Or neither.
</blockquote></details>

Yes, thanks, I should include Win11 in this. I'll update the comment.

Another thing we could do, if we want to guarantee support for older OS's, is add specific versions of Windows to the CI. I can look into this, as well

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: 4 of 12 files reviewed, 3 unresolved discussions (waiting on @andrew-polk, @ermshiperete, and @megahirt)

___
*[`.github/workflows/CI-CD.yml` line 48 at r4](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-Ny69G5YBt6B8VJPyeT_:-NzCWntE0CUUglmbDDOz:b97cwva) ([raw file](https://github.com/sillsdev/l10nsharp/blob/74bc7026629ceb2be42f639f0a1640b948362c7f/.github/workflows/CI-CD.yml#L48)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

We have been discussing a branching strategy for libpalaso [here](https://github.com/sillsdev/libpalaso/issues/1320). I was not planning on applying that to L10nSharp, since other than the 4.6.1 tests and Ubuntu, everything is living happily together. Are you proposing we implement the branching strategy for L10nSharp, as well? That will fundamentally change this PR, and future work on this repo specifically, but it would allow us to more explicitly support the theoretical 4.6.1/Mono userbase.

Maybe these changes would be more appealing if I were actually able to get the ubuntu build to work? This would be regardless of whether we want 4.6.1 on the master branch or an eventual framework branch. I will investigate some more.

(I'm not confident about getting all three sets of tests to run.)
</blockquote></details>

So I was able to get the Ubuntu build reinstated. I had to remove the tests for that OS, since they were failing (Framework 4.6.1 and 4.8), but on master those tests aren't doing anything anyway. They are "succeeding" in 2 seconds with empty output. So it's clearer to remove the build step. We can open a separate issue to add this feature, if desired.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: 4 of 12 files reviewed, 2 unresolved discussions (waiting on @andrew-polk, @ermshiperete, and @megahirt)

___
*[`src/L10NSharp/L10NCultureInfo.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/118#-NxY8L8y35Yjt22poWQ3:-NzCgA6XEH0v7l384nQd:b-frhg1z) ([raw file](https://github.com/sillsdev/l10nsharp/blob/6e530a0d8d8f5ee3289b100e106d1b1228fbb548/src/L10NSharp/L10NCultureInfo.cs#L39)):*
<details><summary><i>Previously, josephmyers wrote…</i></summary><blockquote>

Yes, thanks, I should include Win11 in this. I'll update the comment.

Another thing we could do, if we want to guarantee support for older OS's, is add specific versions of Windows to the CI. I can look into this, as well
</blockquote></details>

After a little research, the oldest windows runner supported by GHA is `windows-2019`. Doesn't really seem super valuable to add that to the list on top of `windows-latest`, so I recommend we close this out.

josephmyers:
Given the numerous changes here surrounding platforms and their tests, I made some super simple tables that lay out who all is running what.

Master:
| | Windows | Ubuntu |
| ------------- | ------------- | ------------- | 
| Framework 4.6.1  | Yes | - | 
| Framework 4.8  | - | - |
| .NET8  | -  | - |

This PR:
| | Windows | Ubuntu |
| ------------- | ------------- | ------------- | 
| Framework 4.6.1  | Yes | - | 
| Framework 4.8  | Yes | - |
| .NET8  | Yes | - |

hahn-kev:
just checked this out, ran the tests and it all seems to work correctly. Can we get this merged in?

hahn-kev (commented):
> Reviewed 1 of 7 files at r1, 3 of 6 files at r3, 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: all files reviewed, 2 unresolved discussions (waiting on @ermshiperete and @megahirt)

andrew-polk (approved):
> Reviewed 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/118)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ermshiperete)

josephmyers:
@hahn-kev I don't think this will solve #116, but it might get you closer.

PR #122, opened 2025-02-12
Use Version.TryParse when parsing version in existing XLIFF file

tombogle:
This will prevent a crash if it is corrupt. See https://jira.sil.org/browse/HT-487

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/122)


src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs
```diff
@@ -201,8 +201,11 @@ private void CreateOrUpdateDefaultXliffFileIfNecessary(
 					File.Delete(DefaultStringFilePath);
 					Console.WriteLine("WARNING - L10NSharp Update deleted corrupted {0}", DefaultStringFilePath);
 				}
-				if (verAttribute != null && new Version(verAttribute.Value) >= new Version(AppVersion ?? "0.0.1"))
+				if (verAttribute != null && Version.TryParse(verAttribute.Value, out var existingVer) &&
+					    existingVer >= new Version(AppVersion ?? "0.0.1"))
```
hahn-kev, line 205: Wouldn't this also throw? Should we do the same thing here?

tombogle: The calling application has control over this, so if it passes a bogus version number, I guess we probably want it to throw an exception. (If we don't, then I think we might end up writing the file with a bad version number.) That said, we could consider catching the exception and wrapping it with a more helpful exception that includes the bogus version string.

tombogle: Done.

tombogle (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/122)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @hahn-kev and @josephmyers)

hahn-kev (approved):

src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs
```diff
@@ -332,11 +339,29 @@ public static void ShowLocalizationDialogBox(string id, IWin32Window owner = nul
...
 		/// </summary>
 		/// ------------------------------------------------------------------------------------
-		public string AppVersion { get; }
+		public string AppVersion
+		{
+			get => _appVersion.ToString();
+			set
+			{
+				Debug.Assert(_appVersion == null,
+					"AppVersion is effectively read-only and should be set only in the context of the constructor.");
```
josephmyers, line 352: If AppVersion should be readonly, why don't we make it so? It's only set from the two constructors, and it'd be easy enough to have the one call the other to share code. Then you don't need the private backing field, either. You could make this validation into a private method, or leave it in the three-line constructor.

`public string AppVersion {get;}`

josephmyers (changes requested):

src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs
tombogle: Good call! Fixed.

josephmyers (approved):
Awesome. Looks good

josephmyers (approved):

PR #125, opened 2025-07-22
Split L10NSharp to remove winforms dependency

aror92:
Split L10NSharp into two parts to separate out the portion that does not require use of Windows Forms from the portion that does.

Additionally, remove System.Drawing dependency from the Windows Forms independent portion, since as of .NET 6 System.Drawing.Common is only supported on Windows.

Remove code related to runtime localization.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125)


aror92:
Made the Winforms > .Windows.Forms namespace change. I just checked again and realized the RootNamespace for L10NSharp tests is .Tests. Updated the folder the match.

Also, I'm realizing that I find it quite helpful to have all the dialogues, buttons, helpers, etc in a separate namespace. I'd vote for changing ".UI" to ".UIComponents" and moving L10NExtender out into the Windows.Forms namespace.

aror92:
*[`src/L10NSharp/LocalizationManagerInternal.cs` line 779 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OWNU0jL33fIQlhBaZtp:-OWVhIZ2-qRF1CPcs-a9:bna7g4d) ([raw file](https://github.com/sillsdev/l10nsharp/blob/4aa2bd0a3022993996494963ff2e25b810b7bebf/src/L10NSharp/LocalizationManagerInternal.cs#L779)):*

Any reason to keep this commented code?
</blockquote></details>

Deleted, thanks for catching this.

andrew-polk (commented):
Along with adding to the ChangeLog, you'll want to add "+semver:major" to one of the commit messages.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-:-OWXv2Gl3-YBBOoMwQuG:bdkr50r)* status: 100 of 114 files reviewed, 6 unresolved discussions (waiting on @aror92)

andrew-polk (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125)* status: 100 of 114 files reviewed, 5 unresolved discussions (waiting on @aror92)

___
*[`src/L10NSharp/Properties/AssemblyInfo.cs` line 13 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OWNP8c_D-ySSO6Os7Qt:-OWXwHfg7-f4N1nbva-H:b-pa37rb) ([raw file](https://github.com/sillsdev/l10nsharp/blob/4aa2bd0a3022993996494963ff2e25b810b7bebf/src/L10NSharp/Properties/AssemblyInfo.cs#L13)):*
<details><summary><i>Previously, aror92 (Ariel Ror.) wrote…</i></summary><blockquote>

Happy for alternate suggestions for how to handle this. The issue is that the Winforms package needs to extend the internal interface and class ILocalizationManagerInternal and LocalizationManagerInternal. The solutions that occur to me are either these need to become public, or we need to see L10NSharp internals.
</blockquote></details>

Yeah... if we were really going to do things the cleanest, I think we could go back to not having the "internal" version at all. But 
1) That's biting off more than is worth it for this project.
2) As soon as did that, someone would want to implement another `TranslationMemory` type, and we would have shot ourselves in the foot.

I'm sure this is probably the best scenario given what we're trying to do.

andrew-polk (commented):
> Reviewed 4 of 101 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125)* status: 104 of 114 files reviewed, 3 unresolved discussions (waiting on @aror92)

___
*[`src/L10NSharpWinforms/LocalizationManagerWinforms.cs` line 58 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OWNRQAa7yoLvjTa5nUQ:-OWY-ijM07W5kT7teRF5:b-2j2fv6) ([raw file](https://github.com/sillsdev/l10nsharp/blob/4aa2bd0a3022993996494963ff2e25b810b7bebf/src/L10NSharpWinforms/LocalizationManagerWinforms.cs#L58)):*

As far as I can tell, applicationIcon is never used now. (Looks like it was just for LanguageChoosingDialog?)

[But see other comment about possibly reinstating use of LanguageChoosingDialog elsewhere.]

If that's true, that might mean a lot of clean-up can happen, starting with the removal of this Create method.
(I haven't fully evaluated that.)
</blockquote></details>

Ok, now that you've reinstated LanguageChoosingDialog, I guess that does use it.
___
*[`src/L10NSharpWinforms/LocalizationManagerWinforms.cs` line 82 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OWNSsM4CADmk2RvB2wC:-OWY-ujR3Htpklw_d0ah:b-7xmg1t) ([raw file](https://github.com/sillsdev/l10nsharp/blob/4aa2bd0a3022993996494963ff2e25b810b7bebf/src/L10NSharpWinforms/LocalizationManagerWinforms.cs#L82)):*

I think the only two usages of this should be using `LocalizationManager.LoadedManagers`.
</blockquote></details>

Never mind. I see you switched it the other way and that was probably more correct.

aror92 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125)* status: 125 of 135 files reviewed, 4 unresolved discussions (waiting on @andrew-polk)

___
*[`src/L10NSharp.Windows.Forms.Tests/TempFile.cs` line 21 at r6](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OWqCY4j4mo_FLtIp6B-:-OWwFaT-A3EY2WkUM18H:b-w5psv2) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e081195cd4764bc43a5cb007fed9204d9ab38863/src/L10NSharp.Windows.Forms.Tests/TempFile.cs#L21)):*

Is there a reason you needed to copy this file to Winforms rather than use the existing one?
</blockquote></details>

Not really. I hadn't added a project reference to L10NSharp.Tests in L10NSharp.Windows.Forms.Tests and wasn't sure whether it made more sense to do that or to copy this file. I initially thought I might only need a small piece of what was in this file but ended up needing all of it for one of the tests. I've referenced the existing one instead and remove the duplication.
___
*[`src/L10NSharp.Windows.Forms.Tests/TestXliff/xliff-core-1.2-strict.xsd` line 1 at r6](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OWqD9RK703_QBTlpPOW:-OWwHUwSBdaxjqXZPSjz:bwczdje) ([raw file](https://github.com/sillsdev/l10nsharp/blob/e081195cd4764bc43a5cb007fed9204d9ab38863/src/L10NSharp.Windows.Forms.Tests/TestXliff/xliff-core-1.2-strict.xsd#L1)):*

I don't understand why the .xsd files were copied.
</blockquote></details>

Thanks for catching that. Needed to copy the test xlfs, accidentally copied the xsds as well.
___
*[`src/L10NSharp.Windows.Forms/XLiffUtils/XliffLocalizationManagerWinforms.cs` line 209 at r3](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OWY4Jop9Ccew3CGayyE:-OWwS6oU5Ni8qjzG-PSd:b9xoxqg) ([raw file](https://github.com/sillsdev/l10nsharp/blob/9e1252feb826e6101194d3cb2d5dd03d6a5bde78/src/L10NSharp.Windows.Forms/XLiffUtils/XliffLocalizationManagerWinforms.cs#L209)):*

Looks like there is some stuff to clean up with the Alt-Shift-Click handler which used to bring up the localization dialog.
(This is also referenced in a label in the sample app and README.)
</blockquote></details>

Thanks! Still need to revise the README, but I think this cs file is cleaned up now.

andrew-polk (commented):
Something isn't working right. When I load the SampleApp with the current code, it doesn't seem to localize the components.
If I add back in the Resources.Icon parameters in the LocalizationManagerWinforms.Create calls in the SampleApp, it does work.
I think the difference is that without the icon parameter, it is actually calling the non-winforms Create method and things don't get set up correctly.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-:-OWwfez-DitkLtymEIwr:bblymv)* status: 136 of 146 files reviewed, 2 unresolved discussions (waiting on @aror92)

aror92 (commented):
Oh, yes, LocalizationManagerWinforms is inheriting the Creates from LocalizationManager. I added a new create to LocalizationManagerWinforms that doesn't expect an icon parameter. There are also two obsolete create methods in LocalizationManager that include a TranslationMemory parameter. Since they're obsolete, I haven't added new create methods to hide those in LocalizationManagerWinforms. Should I? (Or is this a good time to remove those obsolete methods?)

The above should fix this issue if create is called without the icon parameter. I've also added that parameter back to the SampleApp.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-:-OX0WBrM2B0U4TW_bBcn:b-mri6cl)* status: 136 of 146 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

jasonleenaylor:
*[`src/L10NSharp.Windows.Forms/LocalizationManagerWinforms.cs` line 159 at r9](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OXdBF-6-2oxOXzMEy4p:-OXdBF-6-2oxOXzMEy4q:b-nb8469) ([raw file](https://github.com/sillsdev/l10nsharp/blob/1e6a6c15fcee27e62ee55e9ff2752032b54cf53c/src/L10NSharp.Windows.Forms/LocalizationManagerWinforms.cs#L159)):*
> ```Smalltalk
> 
> 		/*/// <summary>
> 		/// True (default) to throw if we try to get a string from a particular manager
> ```

Did these get implemented in a different place? It makes sense that these should be in a non-winforms location. Also it would be better if they weren't statics if we ever consider using this outside of a desktop app situation...but that should probably be saved for a future PR.

tombogle:
> Oh, yes, LocalizationManagerWinforms is inheriting the Creates from LocalizationManager. I added a new create to LocalizationManagerWinforms that doesn't expect an icon parameter. There are also two obsolete create methods in LocalizationManager that include a TranslationMemory parameter. Since they're obsolete, I haven't added new create methods to hide those in LocalizationManagerWinforms. Should I? (Or is this a good time to remove those obsolete methods?)
> 
> The above should fix this issue if create is called without the icon parameter. I've also added that parameter back to the SampleApp.
> 
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-:-OX0WBrM2B0U4TW_bBcn:b-mri6cl)_ status: 136 of 146 files reviewed, 1 unresolved discussion (waiting on @andrew-polk)

Since we're breaking everything, this is probably the ideal time to jettison anything marked as obsolete.

andrew-polk (commented):
agree

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-:-OYmvt1tAbjXkbKoCuJL:b-7tdf7d)* status: 132 of 148 files reviewed, all discussions resolved (waiting on @jasonleenaylor and @tombogle)

aror92 (commented):
Okay, great. I actually went ahead and made that change when I was working on the Changelog.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-:-OZBNwMs50XBGuQblnu9:b8bhpqp)* status: 140 of 148 files reviewed, 7 unresolved discussions (waiting on @andrew-polk, @jasonleenaylor, and @tombogle)

___
*[`src/L10NSharp/LocalizingInfo.cs` line 77 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmyxPqCHqC4TfEDM6i:-OZBYE1rAuGoDJ2Xt-Py:b-s4ccoe) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/LocalizingInfo.cs#L77)):*

Can/should this move to Winforms?
</blockquote></details>

LocalizationCategory is currently used in XliffLocalizedStringCache, but I could look into whether it would be possible to move those usages to XliffLocalizedStringCacheWinforms
___
*[`src/L10NSharp/LocalizingInfo.cs` line 89 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmz-3y8AOiqG-qoa0F:-OZBZ3bW2q-2PD9Nmfm3:bp5jsyo) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/LocalizingInfo.cs#L89)):*

Can/should this move to winforms?
</blockquote></details>

UpdateFields is used in the StringExtractor, some XliffUtils, and LocalizationManagerInternal. I don't think we'd want to move it entirely, because we do at least need to use and update Text and Comment in the non-winforms code.
___
*[`src/L10NSharp/LocalizingInfo.cs` line 102 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmzO9v5p6QtM4-cyg0:-OZBSIl58umQbXNKUs4d:b-clj9hv) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/LocalizingInfo.cs#L102)):*

Much of this class seems pretty specific to winforms stuff.
Maybe it was too hard to tease out its usage from other places in the non-winforms code?
</blockquote></details>

Yeah, it's tricky because LocalizingInfo is used a fair amount in the non-winforms code, especially StringExtractor (this is probably the biggest challenge to moving more of LocalizingInfo to the winforms side; it references all the properties at least once), and to a lesser extent some classes in XliffUtils.
___
*[`src/L10NSharp/LocalizingInfo.cs` line 107 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmz55-4S62ASMzDrFy:-OZBS4CfEthae30oRsuR:b-wlgj7l) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/LocalizingInfo.cs#L107)):*

> _shortcutKeys

Can/should this move to winforms?
</blockquote></details>

Nothing is really done with it here in LocalizingInfo. But its value is set in the StringExtractor's "GetInfoForCallToGetStringMethod" method in the non-winforms side (depending on how many parameters are given). If it seems best to move it, I could make some changes in StringExtractor. We may need a Winforms version of StringExtractor too, then?
___
*[`src/L10NSharp/Properties/AssemblyInfo.cs` line 13 at r2](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OWNP8c_D-ySSO6Os7Qt:-OZBP_42EaPFaKDVRp_i:b-de9sey) ([raw file](https://github.com/sillsdev/l10nsharp/blob/4aa2bd0a3022993996494963ff2e25b810b7bebf/src/L10NSharp/Properties/AssemblyInfo.cs#L13)):*

I think I would favor making them public unless there is some clear and present danger (which seems pretty unlikely for an interface).
</blockquote></details>

  

The LocalizationManagerInternal class handles the active manager list and changes to it, which probably could be public, though I think there may be benefit to keeping it internal. If we do want to make these public, I think it would be better as follow-up work in another PR (along the lines of Andrew's item number 1 above). It would involve combining the current public class LocalizationManager and the internal class LocalizationManagerInternal, as well as the interfaces ILocalizationManager and ILocalizationManagerInternal.
___
*[`src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs` line 22 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmyXw085u4FvRgaPW0:-OZBQLef9yFWNcqw8YnF:b-gdmamf) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs#L22)):*

Any reason to keep these here as comments?
</blockquote></details>

Thanks for catching it; deleted
___
*[`src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs` line 166 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmyd0p3yCnxasoiwg-:-OZBQPpY7fWntKpEBDRh:b-9unw5o) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs#L166)):*

Any reason to keep this?
</blockquote></details>

deleted

andrew-polk (changes requested):
> @andrew-polk reviewed 2 of 93 files at r1, 1 of 9 files at r5, 1 of 7 files at r9, 4 of 4 files at r10.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125)* status: all files reviewed, 5 unresolved discussions (waiting on @aror92)

___
*[`src/L10NSharp/LocalizingInfo.cs` line 107 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmz55-4S62ASMzDrFy:-OZBcijh5LkBqwI9KrW_:b-4ux1ix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/LocalizingInfo.cs#L107)):*
<details><summary><i>Previously, aror92 (Ariel Ror.) wrote…</i></summary><blockquote>

Nothing is really done with it here in LocalizingInfo. But its value is set in the StringExtractor's "GetInfoForCallToGetStringMethod" method in the non-winforms side (depending on how many parameters are given). If it seems best to move it, I could make some changes in StringExtractor. We may need a Winforms version of StringExtractor too, then?
</blockquote></details>

I don't want to expand the scope of this project.
But I also don't want to leave anything in an in-between state which can be resolved fairly easily.

I mostly wanted to raise the questions to make sure they are things you've thought about. You have, so I'm satisfied.
___
*[`src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs` line 22 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmyXw085u4FvRgaPW0:-OZBmkeS7LOM71UPaGIY:b-sjkdfh) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs#L22)):*
<details><summary><i>Previously, aror92 (Ariel Ror.) wrote…</i></summary><blockquote>

Thanks for catching it; deleted
</blockquote></details>

I'm not seeing the deletion.
___
*[`src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs` line 166 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmyd0p3yCnxasoiwg-:-OZBmox09Zys_3tit6Cm:b-up4tv2) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs#L166)):*
<details><summary><i>Previously, aror92 (Ariel Ror.) wrote…</i></summary><blockquote>

deleted
</blockquote></details>

Not seeing the deletion.
___
*[`src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs` line 622 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OZBm1m03_j-cGTdVWph:-OZBm1m03_j-cGTdVWpi:bf7u1mw) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs#L622)):*
> ```Smalltalk
> 		}
> 
> 		/*/// ------------------------------------------------------------------------------------
> ```

Any reason to keep this?
___
*[`src/L10NSharp/XLiffUtils/XliffLocalizedStringCache.cs` line 52 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OZBmL4T4xhCU76OisDc:-OZBmL4T4xhCU76OisDd:b-kzslf8) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/XLiffUtils/XliffLocalizedStringCache.cs#L52)):*
> ```Smalltalk
> 				{
> 					//MessageBox.Show("Error occurred reading localization file:" + Environment.NewLine + e.Message,
> 					//	Application.ProductName, MessageBoxButtons.OK, MessageBoxIcon.Warning);
> ```

Seems like it would be better to at least give a Console.WriteLn or something.
And then there's probably no reason to keep the comment.

aror92 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125)* status: 131 of 148 files reviewed, 5 unresolved discussions (waiting on @andrew-polk)

___
*[`src/L10NSharp/LocalizingInfo.cs` line 99 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmyon64MPIGJzP9SA4:-OZPC4Z34qqV8yeeKRAt:b-bu91xm) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/LocalizingInfo.cs#L99)):*

I think this comment needs updating.
</blockquote></details>

Updated
___
*[`src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs` line 622 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OZBm1m03_j-cGTdVWph:-OZK3wgB8C0ElF4mNuaH:b-eeb23c) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs#L622)):*

Any reason to keep this?
</blockquote></details>

Deleted it.

Also, I realized that ApplyLocalization/ReapplyLocalizations/RefreshTooltips were all being called only on the non-Winforms side, so these null-ops were being called. I removed these null-ops here and from the non-Winforms interface and moved calls of these methods from LocalizationManager to LocalizationManagerWinforms so they access the XliffLocalizationManagerWinforms versions of these methods instead.
___
*[`src/L10NSharp/XLiffUtils/XliffLocalizedStringCache.cs` line 52 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OZBmL4T4xhCU76OisDc:-OZGmH2--tZdXllAUGVm:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp/XLiffUtils/XliffLocalizedStringCache.cs#L52)):*

Seems like it would be better to at least give a Console.WriteLn or something.
And then there's probably no reason to keep the comment.
</blockquote></details>

Done.
___
*[`src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs` line 22 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmyXw085u4FvRgaPW0:-OZGmICR6xazI3QFlfub:bsmhafk) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs#L22)):*

I'm not seeing the deletion.
</blockquote></details>

Sorry, forgot to push the change!
___
*[`src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs` line 166 at r10](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OYmyd0p3yCnxasoiwg-:-OZGmOyp-zzZ1wx6YrtO:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/0c2fa6df67a2a792307a7d08b2b1f9acf09a0deb/src/L10NSharp.Windows.Forms/LocalizingInfoWinforms.cs#L166)):*

Not seeing the deletion.
</blockquote></details>

Done.

andrew-polk (approved):
> @andrew-polk reviewed 17 of 17 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/125)* status: all files reviewed, 1 unresolved discussion (waiting on @aror92)

___
*[`CHANGELOG.md` line 76 at r11](https://reviewable.io/reviews/sillsdev/l10nsharp/125#-OZQC58b3LzuO3zUFZtR:-OZQC58b3LzuO3zUFZtS:bhp1n4v) ([raw file](https://github.com/sillsdev/l10nsharp/blob/8071200de76caf7af71e285df095cac35e5cd69b/CHANGELOG.md#L76)):*
> ```Markdown
> 
>     - The properties ApplicationIcon, ToolTipCtrls, LocalizableComponents and StringCache.
>     - The methods RegisterComponentForLocalizing, GetShortcutKeyFromStringCache, ApplyLocalizationToIlocalizableComponent, ReapplyLocalizationsToAllComponents, RefreshToolips, ApplyLocalization, ApplyLocalizationsToILocalizableComponent, ApplyLocalizationsToControl, ApplyLocalizedToolTipToControl, HandleToolTipRefChanged, HandleToolTipRefDestroyed, ApplyLocalizationsToToolStripItem, ApplyLocalizationToListViewColumnHeader, and ApplyLocalizationToDataGridViewColumn.
> ```

RefreshToolips sounds like a lovely sort of method. Very spring-like.

(you accidentally deleted a t)

PR #130, opened 2026-04-21
Update several dependencies

papeh:
to resolve three of four vulnerable transitive dependencies and to be more up to date.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/130)

### DIFF
--- a/src/CheckOrFixXliff/CheckOrFixXliff.csproj
+++ b/src/CheckOrFixXliff/CheckOrFixXliff.csproj
@@ -16,9 +16,9 @@
     <Reference Include="System.Windows.Forms" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="GitVersion.MsBuild" Version="5.11.1" PrivateAssets="all" />
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.1.1" PrivateAssets="all" />
-    <PackageReference Include="SIL.ReleaseTasks" Version="2.5.0" PrivateAssets="all" />
+    <PackageReference Include="GitVersion.MsBuild" Version="6.7.0" PrivateAssets="all" />
+    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="10.0.201" PrivateAssets="all" />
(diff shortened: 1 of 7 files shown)


tombogle (commented):
> @tombogle reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/130)* status: 2 of 7 files reviewed, 1 unresolved discussion (waiting on papeh).

___
*[`src/L10NSharp/L10NSharp.csproj` line 15 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/130#-OqpPKub5mGh0fj_-6TF:-OqpPKub5mGh0fj_-6TG:b-7vn3ue) ([raw file](https://github.com/sillsdev/l10nsharp/blob/315b14991783011476e850e107777bbd02228a16/src/L10NSharp/L10NSharp.csproj#L15)):*
> ```XML
>   <ItemGroup Condition=" '$(TargetFramework)' == 'net8.0-windows' ">
>     <PackageReference Include="System.ServiceModel.Http" Version="10.0.652802" />
>     <PackageReference Include="System.ServiceModel.Primitives" Version="10.0.652802" />
> ```

The changes to `System.ServiceModel.Http` and `System.ServiceModel.Primitives`, if required, need to be treated as breaking changes. They could mean that downstream clients that target .Net Framework 4.6.2 break (possibly in subtle runtime ways).

These packages are part of the **.NET runtime support libraries**, not standalone features. Major version jumps typically track:

-   new .NET target frameworks (.NET 6 → .NET 10 era)
-   updated runtime contracts
-   removed legacy compat behaviors
-   tighter forward-only assumptions

So this isn’t “new features”; it’s “new baseline expectations”.

Even if L10nSharp still targets .NET Standard / .NET Framework, upgrading these packages tends to _silently raise the minimum viable runtime stack_.

That means downstream consumers may suddenly hit:

-   runtime load failures on older .NET Framework apps
-   missing type exceptions in service model components
-   subtle serialization differences

PR #146, opened 2026-06-05
Validate l10n ID is non-empty to prevent crash on empty string ID

### LINKED ISSUE #104: Empty string for l10n ID causes crash
Adding 
`LocalizationManager.GetString("", "Some String");`
in the client causes a 0-length .xlf file to be generated which crashes the application on launch (at least in Bloom).

Clearly, this is errant code, but we should handle it more gracefully.

imnasnainaec:
Closes #104

Passing a null, empty, or whitespace string as the `id` argument to `GetString`/`GetDynamicString` produced a malformed or zero-length `.xlf` file that crashed the application on the next launch.

The fix is in `XliffTransUnitUpdater.Update`: a guard (mirroring the existing `LangId` check) now exits early when `id` is null/empty/whitespace, preventing the corrupt entry from ever being written. The public API methods (`GetDynamicString`, `GetDynamicStringOrEnglish`, `GetString`) return the English fallback text when called with a blank ID rather than attempting a cache lookup or write, so this is a non-breaking fix.

**Note on `StrictInitializationMode`:** the early-return in `GetString` bypasses the `StrictInitializationMode` check inside `GetStringFromAnyLocalizationManager`. An app with `StrictInitializationMode = true` and no managers loaded will now get English fallback text instead of `InvalidOperationException` when a blank ID is passed. This seems like the right behavior — throwing about missing managers when the real problem is a blank ID would be the wrong signal — but it is an observable change in that scenario.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/146)

Devin review: https://app.devin.ai/review/sillsdev/l10nsharp/pull/146

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [L10NSharp] Removed emailForSubmissions parameter from LocalizationManager.Create. Since the localization dialog was jettisoned, it no longer makes sense to store this information on the localization manager.
 - [L10NSharp.Windows.Forms] Removed emailForSubmissions parameter (8th parameter) from LocalizationManagerWinforms.Create. Since the localization dialog was jettisoned, it no longer makes sense to store this information on the localization manager.
 - [L10NSharp] Replaced the .NET 8.0 target with .NET Standard 2.0 for broader compatibility.
+- [L10NSharp] `GetDynamicString`, `GetDynamicStringOrEnglish`, and `GetString` now return the English fallback text immediately when called with a null, empty, or whitespace string ID, rather than attempting a cache lookup or write.
  
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [L10NSharp] Removed emailForSubmissions parameter from LocalizationManager.Create. Since the localization dialog was jettisoned, it no longer makes sense to store this information on the localization manager.
 - [L10NSharp.Windows.Forms] Removed emailForSubmissions parameter (8th parameter) from LocalizationManagerWinforms.Create. Since the localization dialog was jettisoned, it no longer makes sense to store this information on the localization manager.
 - [L10NSharp] Replaced the .NET 8.0 target with .NET Standard 2.0 for broader compatibility.
+- [L10NSharp] BREAKING CHANGE: `LocalizationManager.GetString`, `GetDynamicString`, and `GetDynamicStringOrEnglish` now throw `ArgumentException` when called with a null or empty string ID. Previously, empty IDs were silently accepted and could produce a malformed XLIFF file that crashed on next launch.
  
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


andrew-polk (changes requested):
If we're truly treating this as a breaking change, it needs `semver:major` in the commit message.

Initially, I was going to approve the general idea, but I'm getting more and more hesitant.
I wouldn't be at all surprised if Bloom (or some other program) is accidentally making use of the fallback behavior and sending null or empty sometimes.

Can we solve instead solve the underlying problem while maintaining the immediate behavior of falling back to english?

I'm not sure what can cause the corrupt file, but can we just exit early, returning the English or ID instead of throwing, for example?

> @andrew-polk reviewed 4 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/146#-:-Ow4681p-1KV9LXOR8qf:bsbgx7d)* status: all files reviewed, 1 unresolved discussion (waiting on imnasnainaec).

___
*[`CHANGELOG.md` line 35 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/146#-Ow46dHy7cHg5HyQFlbp:-Ow46dHy7cHg5HyQFlbq:b-99tyrv) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d58956037112b1973d90462a374ba63e38a23cef/CHANGELOG.md#L35)):*
> ```Markdown
> - [L10NSharp.Windows.Forms] Removed emailForSubmissions parameter (8th parameter) from LocalizationManagerWinforms.Create. Since the localization dialog was jettisoned, it no longer makes sense to store this information on the localization manager.
> - [L10NSharp] Replaced the .NET 8.0 target with .NET Standard 2.0 for broader compatibility.
> - [L10NSharp] BREAKING CHANGE: `LocalizationManager.GetString`, `GetDynamicString`, and `GetDynamicStringOrEnglish` now throw `ArgumentException` when called with a null or empty string ID. Previously, empty IDs were silently accepted and could produce a malformed XLIFF file that crashed on next launch.
> ```

This says null or empty, but the actual check is whitespace.

imnasnainaec (commented):
Went for the English/empty fallback instead of the throw.

> @imnasnainaec made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/146#-:-OwIktZU-fW17fV0i_o0:b7pd8eb)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on andrew-polk).

___
*[`CHANGELOG.md` line 35 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/146#-Ow46dHy7cHg5HyQFlbp:-OwIm2XB4MuexA-U0LZh:b-896fix) ([raw file](https://github.com/sillsdev/l10nsharp/blob/d58956037112b1973d90462a374ba63e38a23cef/CHANGELOG.md#L35)):*

This says null or empty, but the actual check is whitespace.
</blockquote></details>

Done.

andrew-polk (approved):
> @andrew-polk reviewed 4 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/146)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on imnasnainaec).

PR #153, opened 2026-06-08
Fix NRE in XLiffBody.AddTransUnit when tu.Source is null

imnasnainaec:
## Summary
- Adds null guard on `tu.Source` in `AddTransUnit` so a null Source doesn't throw when Target is also null
- Adds null guard on `tu.Source` in the `NumberTranslated` property for the same case
- Guards `tu.Source`/`tuOld.Source` in `MergeXliffDocuments` against the same NRE when a baseline XLIFF unit has no source variant; a blank old source is still counted as changed but no longer emits an empty `OLD TEXT` note
- Simplifies the null checks across `AddTransUnit`, `NumberTranslated`, and `NumberApproved` to use null-conditional (`?.`) access. As part of this, the `NumberTranslated` "differs from source" heuristic now requires a non-blank source (`IsNullOrWhiteSpace`), so a unit with a blank source no longer trivially counts as translated. `AddTransUnit` keeps `!= null` since an empty/whitespace value is still a legitimate stored translation.

Closes #149

Devin review: https://app.devin.ai/review/sillsdev/l10nsharp/pull/153

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/153)


tombogle (approved):
> @tombogle reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/153)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on imnasnainaec).

PR #157, opened 2026-07-09
migrate core library to use nullable reference types

hahn-kev:
After seeing some fixes from @imnasnainaec, I figured I'd try my hand at enabling NRT and fixing some issues that came up. 

There's currently 4 warnings still in place, I wasn't sure what the right call would be, so we can leave them in or come up with a fix together. They are:
```
    src\L10NSharp\XLiffUtils\XLiffBody.cs(173,22): warning CS8604: Possible null reference argument for parameter 'key' in 'string ConcurrentDictionary<string, string>.this[string key]'.
    src\L10NSharp\XLiffUtils\XLiffBody.cs(175,22): warning CS8604: Possible null reference argument for parameter 'key' in 'string ConcurrentDictionary<string, string>.this[string key]'.
    src\L10NSharp\XLiffUtils\XLiffBody.cs(201,21): warning CS8604: Possible null reference argument for parameter 'key' in 'string ConcurrentDictionary<string, string>.this[string key]'.
    src\L10NSharp\XLiffUtils\XliffLocalizationManager.cs(126,12): warning CS8618: Non-nullable property 'DefaultStringFilePath' must contain a non-null value when exiting constructor. Consider adding the 'required' modifier or declaring the property as nullable.
```

There's a lot of code that used `XLiffTransUnit.Id` with the assumption that it's not null, however it can be, I've fixed many of those.

Some places I've added checks like `if x is null throw new exception` this is only in cases where we would have thrown anyway when accessing a property on x later down, so while this may not fix a bug, it does give much better feedback as to what was null.

There's a few places where code may have been returning null, which are now returning an empty string, these were changed due to the method already returning an empty string elsewhere, with the assumption that the method was written to never return null, and these cases were missed.

One change I did make is to constrain `ILocalizationManagerInternal<T>` T to only be a class, this may effect consumers, but I don't think that's possible as the only implementation of this internal interface is a T of `XLiffDocument`, and I think if anything else were to be used there would be a crash at runtime, so I don't think it's a problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/157)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - [L10NSharp.Windows.Forms] Removed emailForSubmissions parameter (8th parameter) from LocalizationManagerWinforms.Create. Since the localization dialog was jettisoned, it no longer makes sense to store this information on the localization manager.
 - [L10NSharp] Replaced the .NET 8.0 target with .NET Standard 2.0 for broader compatibility.
 - [L10NSharp] `GetDynamicString`, `GetDynamicStringOrEnglish`, and `GetString` now return the English fallback text immediately when called with a null, empty, or whitespace string ID, rather than attempting a cache lookup or write.
- 
+- [L10NSharp] Enabled nullable reference types on the core project with full CS86xx warning cleanup across all layers; `ILocalizedStringCache` and `ILocalizationManagerInternal` gained nullable annotations that may require source updates for NRT consumers.
+
 ### Fixed
 
 - [L10NSharp.Windows.Forms] Restored project-local Resources support for `FallbackLanguagesDlgBase` button images (`Move`, `Move_up`, and `Move_down`).
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -25,7 +25,7 @@ See full changelog at https://github.com/sillsdev/l10nsharp/blob/master/CHANGELO
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
     <EnableWindowsTargeting>true</EnableWindowsTargeting>
-    <LangVersion>8.0</LangVersion>
+    <LangVersion>9.0</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <!-- Without this line some projects fail to build on TC with "error : SourceRoot items
--- a/src/L10NSharp.Tests/LocalizationManagerTestsBase.cs
+++ b/src/L10NSharp.Tests/LocalizationManagerTestsBase.cs
@@ -11,7 +11,7 @@ using NUnit.Framework;
 
 namespace L10NSharp.Tests
 {
-	public abstract class LocalizationManagerTestsBase<T> where T : IDocument
+	public abstract class LocalizationManagerTestsBase<T> where T : class, IDocument
 	{
 		protected const string AppId = "test";
 		protected const string AppName = "unit test";
--- a/src/L10NSharp.Windows.Forms/ILocalizationManagerInternalWinforms.cs
+++ b/src/L10NSharp.Windows.Forms/ILocalizationManagerInternalWinforms.cs
@@ -20,7 +20,7 @@ namespace L10NSharp.Windows.Forms
 			Action<ILocalizationManagerInternalWinforms, LocalizingInfoWinforms> successAction);
 	}
 
-	internal interface ILocalizationManagerInternalWinforms<T> : ILocalizationManagerInternalWinforms, ILocalizationManagerInternal<T>
+	internal interface ILocalizationManagerInternalWinforms<T> : ILocalizationManagerInternalWinforms, ILocalizationManagerInternal<T> where T : class
 	{
 		new ILocalizedStringCacheWinforms<T> StringCache { get; }
 	}
--- a/src/L10NSharp.Windows.Forms/ILocalizedStringCacheWinforms.cs
+++ b/src/L10NSharp.Windows.Forms/ILocalizedStringCacheWinforms.cs
@@ -4,7 +4,7 @@ using L10NSharp.Windows.Forms.UIComponents;
 
 namespace L10NSharp.Windows.Forms
 {
-	internal interface ILocalizedStringCacheWinforms<T> : L10NSharp.ILocalizedStringCache<T>
+	internal interface ILocalizedStringCacheWinforms<T> : L10NSharp.ILocalizedStringCache<T> where T : class
 	{
 		List<LocTreeNode<T>> LeafNodeList { get; }
 		Keys GetShortcutKeys(string langId, string id);
--- a/src/L10NSharp.Windows.Forms/LocalizationManagerInternalWinforms.cs
+++ b/src/L10NSharp.Windows.Forms/LocalizationManagerInternalWinforms.cs
@@ -13,7 +13,7 @@ using L10NSharp.Windows.Forms.UIComponents;
 
 namespace L10NSharp.Windows.Forms
 {
-	internal class LocalizationManagerInternalWinforms<T> : LocalizationManagerInternal<T>
+	internal class LocalizationManagerInternalWinforms<T> : LocalizationManagerInternal<T> where T : class
 	{
 
 		/// <summary>
--- a/src/L10NSharp.Windows.Forms/UIComponents/FallbackLanguagesDlg.cs
+++ b/src/L10NSharp.Windows.Forms/UIComponents/FallbackLanguagesDlg.cs
@@ -3,7 +3,7 @@ using System.Linq;
 
 namespace L10NSharp.Windows.Forms.UIComponents
 {
-	public class FallbackLanguagesDlg<T> : FallbackLanguagesDlgBase
+	public class FallbackLanguagesDlg<T> : FallbackLanguagesDlgBase where T : class
 	{
 		public FallbackLanguagesDlg()
 		{
--- a/src/L10NSharp.Windows.Forms/UIComponents/LocTreeNode.cs
+++ b/src/L10NSharp.Windows.Forms/UIComponents/LocTreeNode.cs
@@ -5,7 +5,7 @@ using System.Windows.Forms;
 namespace L10NSharp.Windows.Forms.UIComponents
 {
 	/// ----------------------------------------------------------------------------------------
-	internal class LocTreeNode<T> : TreeNode
+	internal class LocTreeNode<T> : TreeNode where T : class
 	{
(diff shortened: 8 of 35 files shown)


imnasnainaec (changes requested):
> @imnasnainaec reviewed 8 files and all commit messages, and made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/l10nsharp/157)* status: 8 of 35 files reviewed, 6 unresolved discussions (waiting on hahn-kev).

___
*[`src/L10NSharp/LocalizationManagerInternal.cs` line 569 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/157#-Ox5tnPnCVoX_zOyGbmD:-Ox5tnPnCVoX_zOyGbmE:berljnf) ([raw file](https://github.com/sillsdev/l10nsharp/blob/f14ab15aabe710f560979cf05f957e891870e3bf/src/L10NSharp/LocalizationManagerInternal.cs#L569)):*
> ```Smalltalk
> 		internal static string? MapToExistingLanguageIfPossible(string? langId)
> 		{
> 			if (langId is null || string.IsNullOrEmpty(langId))
> ```

Devin bug:

src/L10NSharp/LocalizationManagerInternal.cs:R569-570

**Non-null empty string input can produce null output, violating the declared nullability contract**

The language-mapping method returns null for a non-null empty-string input (`return null` at `src/L10NSharp/LocalizationManagerInternal.cs:570` when `string.IsNullOrEmpty` is true), but its `NotNullIfNotNull` attribute promises a non-null return whenever the argument is non-null, so callers may use the result without a null check.

Impact: If a caller passes an empty (but non-null) language ID, the returned value is unexpectedly null, which could cause a null-reference crash in downstream lookups.

<details><summary>The guard conflates null and empty but the attribute only covers null</summary>

At `src/L10NSharp/LocalizationManagerInternal.cs:566-570`:
```csharp
[return: NotNullIfNotNull("langId")]
internal static string? MapToExistingLanguageIfPossible(string? langId)
{
    if (langId is null || string.IsNullOrEmpty(langId))
        return null;
```

`NotNullIfNotNull("langId")` tells the compiler: "if `langId` is not null, the return value is not null." But `string.IsNullOrEmpty("")` is true, so passing `""` (non-null) returns `null`, violating the contract. Callers like `MapToExistingLanguageOrAddMapping` at `src/L10NSharp/LocalizationManagerInternal.cs:659-662` pass a non-null `langId` and use the result without null checks, trusting the attribute.

The fix should either return `langId` (i.e., `""`) instead of `null` for the empty-string case, or remove the `NotNullIfNotNull` attribute.

</details>
___
*[`src/L10NSharp/XLiffUtils/XLiffTransUnit.cs` line 117 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/157#-Ox5tSYEDaCdxzCbmr2g:-Ox5tSYGBFEd1y7JRjOv:b5nmmcv) ([raw file](https://github.com/sillsdev/l10nsharp/blob/f14ab15aabe710f560979cf05f957e891870e3bf/src/L10NSharp/XLiffUtils/XLiffTransUnit.cs#L117)):*
> ```Smalltalk
> 		[XmlIgnore]
>         [MemberNotNullWhen(false, nameof(Id), nameof(Source), nameof(Target))]
> ```

Devin bug:

src/L10NSharp/XLiffUtils/XLiffTransUnit.cs:R117-119

**Incorrect nullability guarantee on translation unit emptiness check can hide null-reference crashes**

The emptiness check on a translation unit incorrectly promises that three properties are non-null whenever the unit is non-empty (`[MemberNotNullWhen(false, ...)]` at `src/L10NSharp/XLiffUtils/XLiffTransUnit.cs:117`), so the compiler silently suppresses null warnings for code that accesses those properties after the check.

Impact: Code that trusts the compiler's null-safety analysis after checking non-emptiness may dereference null `Target`, `Source`, or `Id` at runtime, causing a crash.

<details><summary>The IsEmpty logic uses AND, but the attribute requires all-or-nothing non-nullness</summary>

`IsEmpty` is defined as:
```
IsNullOrEmpty(Id) && Notes.Count == 0 && Source == null && Target == null
```
It returns `false` when ANY single condition is false. For example, a `XLiffTransUnit` with a valid `Id` ("foo") but `Target == null` has `IsEmpty == false`, yet `Target` is still null. The `MemberNotNullWhen(false, nameof(Id), nameof(Source), nameof(Target))` attribute tells the compiler that ALL three are non-null whenever `IsEmpty` is false, which is incorrect.

In practice, `Target` is commonly null for non-empty translation units (it is only set for translated strings — see the constructor at `src/L10NSharp/XLiffUtils/XLiffTransUnit.cs:19-20` which sets `Target = null`). Any future code (or current code in callers) that checks `!tu.IsEmpty` and then accesses `tu.Target.Value` without a null check will crash, and the compiler will not warn about it.

</details>
___
*[`src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs` line 304 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/157#-Ox5uJyZ0Odvo1hV0Ny0:-Ox5uJyZ0Odvo1hV0Ny1:bl3p0hu) ([raw file](https://github.com/sillsdev/l10nsharp/blob/f14ab15aabe710f560979cf05f957e891870e3bf/src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs#L304)):*
> ```Smalltalk
> 		internal string DefaultStringFilePath { get; }
> 
> 		internal string DefaultInstalledStringFilePath =>
> ```

Devin flag:

src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs:R304-307

**DefaultInstalledStringFilePath passes nullable field to Path.Combine without null guard**

`_installedXliffFileFolder` was changed from `string` to `string?` at line 20, but `DefaultInstalledStringFilePath` at `src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs:304-307` still passes it directly to `Path.Combine`, which would throw `ArgumentNullException` if null. This is currently safe because `DefaultInstalledStringFilePath` is only accessed from `XliffLocalizedStringCache.MergeXliffFilesIntoCache` (at `src/L10NSharp/XLiffUtils/XliffLocalizedStringCache.cs:135`), which is only called from the full constructor that always sets `_installedXliffFileFolder` to a non-null value. However, the type system no longer enforces this invariant, making it fragile.
___
*[`src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs` line 310 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/157#-Ox5uXN24_Cpd-Z1aXn-:-Ox5uXN24_Cpd-Z1aXn0:b268790) ([raw file](https://github.com/sillsdev/l10nsharp/blob/f14ab15aabe710f560979cf05f957e891870e3bf/src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs#L310)):*
> ```Smalltalk
> 
> 		public ILocalizedStringCache<XLiffDocument> StringCache { get; } = null!;
> ```

Devin flag:

src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs:R310

**StringCache property uses null-forgiving initializer, relying on constructor ordering**

The `StringCache` property at `src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs:310` is initialized as `= null!`, meaning the compiler trusts it will be set before use. It is indeed set in the full constructor at line 107 (`StringCache = new XliffLocalizedStringCache(this)`), but the minimal constructor at line 126 does NOT set it. If the minimal constructor is used directly (it's `internal`), any access to `StringCache` would throw a `NullReferenceException`. Currently the minimal constructor is only called via the `: this(...)` chain from the full constructor, so this is safe, but it's fragile — a future internal caller using the minimal constructor would hit a null deref with no compiler warning.
___
*[`src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs` line 607 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/157#-Ox5urs69nyjiME8_VBj:-Ox5urs69nyjiME8_VBk:b-odc0cp) ([raw file](https://github.com/sillsdev/l10nsharp/blob/f14ab15aabe710f560979cf05f957e891870e3bf/src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs#L607)):*
> ```Smalltalk
> 				if (xliffOld != null)
> 				{
> 					var tuOld = xliffOld.File.Body.GetTransUnitForId(tu.Id!);
> ```

Devin flag:

src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs:R603-675

**Null-forgiving operators on tu.Id throughout MergeXliffDocuments assume IDs are always present**

Throughout `MergeXliffDocuments` in `src/L10NSharp/XLiffUtils/XliffLocalizationManager.cs`, `tu.Id!` is used extensively (lines 603, 607, 637, 640, 644, 656, 663, 667, 673, 675). Since `XLiffTransUnit.Id` is now `string?`, these null-forgiving operators suppress warnings. The assumption is safe for translation units read from XLIFF files (the XML deserializer populates the `id` attribute), and for units that passed through `AddTransUnitRaw` (which assigns an ID if missing). However, the widespread use of `!` means a malformed XLIFF file with a missing `id` attribute could cause `NullReferenceException` deep in the merge logic rather than a clear error at parse time.
___
*[`src/L10NSharp/XLiffUtils/XliffLocalizedStringCache.cs` line 55 at r1](https://reviewable.io/reviews/sillsdev/l10nsharp/157#-Ox5vCA-7Hacp3eK8Bd2:-Ox5vCA0DwgmcJ3i6HiP:b-rjlnhl) ([raw file](https://github.com/sillsdev/l10nsharp/blob/f14ab15aabe710f560979cf05f957e891870e3bf/src/L10NSharp/XLiffUtils/XliffLocalizedStringCache.cs#L55)):*
> ```Smalltalk
> 					Console.WriteLine(e.Message);
> 					LocalizationManager.SetUILanguage(LocalizationManager.kDefaultLang);
> 					if (DefaultXliffDocument is null)
> ```

Devin flag:

src/L10NSharp/XLiffUtils/XliffLocalizedStringCache.cs:R50-57

**XliffLocalizedStringCache constructor may leave DefaultXliffDocument null on error path**

In `src/L10NSharp/XLiffUtils/XliffLocalizedStringCache.cs:50-57`, when `loadAvailableXliffFiles` is true and `MergeXliffFilesIntoCache` throws, the catch block calls `LocalizationManager.SetUILanguage(kDefaultLang)` and then checks `if (DefaultXliffDocument is null) throw`. This is good defensive coding added in this PR. However, the `_tuUpdater` initialization at line 62 and the `DefaultXliffDocument.File.AmpersandReplacement` access at line 64 both assume `DefaultXliffDocument` is non-null. If the exception path somehow continues (e.g., if the null check is removed in the future), these would crash. The `[MemberNotNull]` attribute on `MergeXliffFilesIntoCache` at line 127 correctly enforces this for the normal path.

